PR #2, opened 2016-06-25
Add openQA install testcase using current openQA on current Tumbleweed

okurz:
Test triggered with:
 - INSTALL=1
 - UPDATE=0
 - INSTALL_ONLY=1
 - HDD_1=opensuse-Tumbleweed-x86_64-20160623-gnome@64bit.qcow2

Local verification run: http://lord.arch/tests/1717

Needles needles in https://github.com/os-autoinst/os-autoinst-needles-openQA/pull/2

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -1,0 +1,16 @@
+package utils;
+
+use base Exporter;
+use Exporter;
+use strict;
+use testapi;
+
+our @EXPORT = qw/wait_for_desktop/;
+
+sub wait_for_desktop {
+    check_screen [qw/boot-menu openqa-desktop/];
+    send_key 'ret' if match_has_tag 'boot-menu';
+    assert_screen "openqa-desktop", 500;
+}
+
+1;
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -10,6 +10,11 @@ our @EXPORT = qw/wait_for_desktop/;
 sub wait_for_desktop {
     check_screen [qw/boot-menu openqa-desktop/];
     send_key 'ret' if match_has_tag 'boot-menu';
+    if (match_has_tag('openqa-desktop-locked')) {
+        send_key 'esc';
+        wait_still_screen(1);
+        type_string $testapi::pasword . "\n";
+    }
     assert_screen "openqa-desktop", 500;
 }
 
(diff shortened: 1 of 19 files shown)


coolo:
I merged it as it's surely better than having outdated tests, but we can't deploy this easily as the latest working TW version is unknown. For this purpose it would be better if the openqa tests were part of distri-opensuse

coolo:
But to test openQA it's properly easier to rebase your tests on the predefined 42.1 images.

coolo:
anyway, for a one off: https://openqa.opensuse.org/tests/215887

Clone_job is supposed to be called as geekotest btw

lib/utils.pm
```diff
@@ -0,0 +1,21 @@
+package utils;
+
+use base Exporter;
+use Exporter;
```
aaannz: Is it required to inherit from Exporter and then import it?

coolo: No, use base is require + push(@ISA) - but this is just a copy of the one in os-autoinst-distri-opensuse

coolo: (hi Isa :)

tests/update/zypper_up.pm
```diff
@@ -1,23 +1,26 @@
...
 
 sub run {
-    assert_screen "openqa-desktop", 500;
+    wait_for_desktop;
     send_key "ctrl-alt-f2";
     assert_screen "inst-console";
     type_string "root\n";
-    assert_screen "password-prompt", 10;
+    assert_screen "password-prompt";
     type_string "1\n";
```
aaannz: I'm confused here. Don't we have select_console for this? And isn't the password different for 13.2 and Tumbleweed base?

okurz: There is "select_console" but I first have to setup all consoles in lib/susedistribution.pm or clean it up more. Remember, we are in os-autoinst-distri-openQA, not os-autoinst-distri-opensuse.

The password "1" is the password that has been used in the pre-generated 13.2 image that has an already installed openQA instance included.

okurz:
> I merged it as it's surely better than having outdated tests, but we can't deploy this easily as the latest working TW version is unknown. For this purpose it would be better if the openqa tests were part of distri-opensuse
> But to test openQA it's properly easier to rebase your tests on the predefined 42.1 images.


nice. I intend to read dynamically the most recent Tumbleweed version and trigger accordingly. os-autoinst-distri-openQA could be our next to "default" testing suite to find out if os-autoinst and/or openQA are too much intermangled with SUSE distribution testing.

PR #5, opened 2016-09-18
Create README.md

okurz:
@sysrich as you are the only other contributor, do you agree with the MIT license or suggest a different one?

---

This change is [Reviewable](https://reviewable.io/reviews/os-autoinst/os-autoinst-distri-openqa/5)


okurz:
*bump*

PR #6, opened 2016-11-09
test_distribution: Downloading from external sources can take a very long time

okurz:
See https://openqa.opensuse.org/tests/302354#step/test_distribution/1


okurz:
*bump*

PR #9, opened 2016-12-03
Add post fail diagnosis for not running job

okurz:
Local verification run: http://lord.arch/tests/5462

### DIFF
--- a/tests/osautoinst/test_running.pm
+++ b/tests/osautoinst/test_running.pm
@@ -8,8 +8,21 @@ sub run {
     type_string "clear\n";
 }
 
+sub get_log {
+    my ($cmd, $name) = @_;
+    my $ret = script_run "$cmd | tee $name";
+    upload_logs $name if $ret;
+}
+
 sub post_fail_hook {
-    assert_script_run 'openqa-client jobs';
+    script_run 'openqa-client jobs';
+    save_screenshot;
+    get_log 'journalctl --pager-end --no-tail --no-pager -u apache2 -u openqa-scheduler -u openqa-websockets -u openqa-webui -u openqa-worker@1' => 'openqa_services.log';
+    get_log 'cat /var/lib/openqa/pool/1/autoinst-log.txt /var/lib/openqa/testresults/*/*/autoinst-log.txt' => 'autoinst-log.txt';
+    script_run 'lsmod | grep kvm';
+    save_screenshot;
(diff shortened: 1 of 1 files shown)


okurz:
@asmorodskyi you were right with the hint regarding the qemu cpu type. It works now, see http://lord.arch/tests/5462 . I guess I started the test with a different worker class in the mean time which set the cpu type explicitly to `qemu64`. Now I set `QEMUCPU=host` and it works.

tests/osautoinst/test_running.pm
```diff
@@ -8,8 +8,21 @@ sub run {
...
+    my ($cmd, $name) = @_;
+    my $ret = script_run "$cmd | tee $name";
+    upload_logs $name if $ret;
+}
+
 sub post_fail_hook {
-    assert_script_run 'openqa-client jobs';
+    script_run 'openqa-client jobs';
+    save_screenshot;
+    get_log 'journalctl --pager-end --no-tail --no-pager -u apache2 -u openqa-scheduler -u openqa-websockets -u openqa-webui -u openqa-worker@1' => 'openqa_services.log';
```
Martchus: This will not work if someone didn't start the worker using systemctl or not as `openqa-worker@1`. But since this is not the production case I assume it is not a problem.

Martchus:
Did you test it? I mean the verification run didn't fail but as I understand it this hook is supposed to only run after failure.

okurz:
Yes, I tested: http://lord.arch/tests/5436#step/dashboard/12
Since then I only moved it to another module. Should be fine :-)

tests/osautoinst/test_running.pm
```diff
@@ -8,8 +8,21 @@ sub run {
     type_string "clear\n";
 }
 
+sub get_log {
```
asmorodskyi: It's looks like something more generic than just one module function. Shouldn't we have something like this in utils.pm ? If not we need to move this one there

okurz: 1. there is no `utils.pm` in this test distribution
2. as soon as we want to use it elsewhere it should be moved to a more generic file. I am judging with [no functionality is added early](http://www.extremeprogramming.org/rules/early.html) here

asmorodskyi: ok

asmorodskyi:
LGTM

PR #10, opened 2017-01-12
Collect openQA logs in all tests

okurz:
Verification run: http://lord.arch/tests/5540

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -1,8 +1,23 @@
 package openQAcoretest;
 use base "basetest";
+use testapi;
+
+
+sub get_log {
+    my ($cmd, $name) = @_;
+    my $ret = script_run "$cmd | tee $name";
+    upload_logs($name) if $ret;
+}
+
+sub post_fail_hook {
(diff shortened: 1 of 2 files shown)


okurz:
ok, that did not actually work but I can see that the problem is OOM so let's add a bit of RAM.

okurz:
fixed

lib/openQAcoretest.pm
```diff
@@ -1,8 +1,24 @@
 package openQAcoretest;
 use base "basetest";
+use testapi;
 
-# All steps belonging to core openQA functionality are 'fatal'. by default
 
+sub get_log {
+    my ($cmd, $name) = @_;
+    my $ret = script_run "$cmd | tee $name";
+    upload_logs $name if $ret;
```
coolo: as you're asking for my review: I suggest to use () if you use post-if

okurz: sure, feedback appreciated, done

okurz:
no, wait. failed

okurz:
fixed

PR #12, opened 2017-02-02
Shutdown machine to allow image uploading

okurz:
See https://openqa.opensuse.org/t345732

Verification run:  http://lord.arch/tests/5616


okurz:
testing, waitfor http://lord.arch/tests/

PR #13, opened 2017-02-14
Accept longer shutdown time

okurz:
Also try to debug shutdown problems in case of stuck on plymouth.

See https://openqa.opensuse.org/t354252


dzedro:
merged but post_fail_hook will not work in this case as test will be already killed

okurz:
> merged but post_fail_hook will not work in this case as test will be already killed

I think not in all cases, e.g. see https://openqa.opensuse.org/tests/354252#step/shutdown/3 . looks like the machine is still up.

PR #17, opened 2017-03-21
Add test for logout

nicksinger:
To prevent tests failing due to still existing session from previous
runs.

Prove run: http://10.162.31.103/tests/83
Needles: https://github.com/os-autoinst/os-autoinst-needles-openQA/pull/8

### DIFF
--- a/main.pm
+++ b/main.pm
@@ -46,6 +46,7 @@ sub load_osautoinst_tests() {
 sub load_openQA_tests() {
     loadtest "openQA/dashboard.pm";
     loadtest "openQA/login.pm";
+    loadtest "openQA/logout.pm";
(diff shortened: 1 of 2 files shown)


nicksinger:
The reason why i created this, is to get a clean environment for tests that use the same test set on top of the HDD which gets populated by this test. In other words, it allows this test to be recursive on its own - which is quiet necessary since it is so versatile.
Anyhow: I agree on the point that it can break tests running later one like "admin" so I'll change that. But i hope my motivation for this change is a little bit more reasonable now.

okurz:
well, show me how you make use of it first, please :-)

coolo:
are we done here? :)

okurz:
I guess now we are done

PR #19, opened 2017-05-14
Prevent sporadic too-early read of nohup.out

okurz:
See https://openqa.opensuse.org/tests/403504#step/openqa_webui/45


okurz:
@asmorodskyi you are getting very good at code-reviewing. Want to review os-autoinst-distri-openQA as well?

PR #21, opened 2017-06-14
Add access to openQA repos for 42.2 and 42.3

michelmno:
Note that at time of this patch there is not yet
repos in OBS for openQA Leap 42.3
https://build.opensuse.org/project/show/devel:openQA
https://build.opensuse.org/project/subprojects/devel:openQA

Signed-off-by: Michel Normand <normand@linux.vnet.ibm.com>


okurz:
so repos should be added correctly but we don't have tests that try this and the repos don't exist (yet). I asked the OBS project admins to please create it. With https://build.opensuse.org/request/show/504637 and https://build.opensuse.org/request/show/504636 we might get all necessary requirements into openSUSE:Leap:42.3 so we would not need an additional repo for dependencies.

PR #25, opened 2017-07-04
Use install_minimalx test for PowerPC (version 2)

michelmno:
because there is no memtest for this architecture.

Signed-off-by: Michel Normand <normand@linux.vnet.ibm.com>

patch validated in my openqa local instance with ppc64le (I do not have x86_64)
related video at https://michelmno.fedorapeople.org/tmp3/video.ogv


okurz:
great work, appreciated!

PR #27, opened 2017-07-07
Add workaround for Dashboard Test

DrMullings:
Implemented a workaround for poo#19798 where
Firefox takes too long to start in the dashboard
test and it then fails.
    
Now waits 180 seconds, records a soft_failure
if ff did not start and then waits another
360 seconds

Verification:
http://pinky.arch.suse.de/tests/40#

Issues:
https://progress.opensuse.org/issues/19798

### DIFF
--- a/tests/openQA/dashboard.pm
+++ b/tests/openQA/dashboard.pm
@@ -9,7 +9,11 @@ sub run {
     ensure_unlocked_desktop();
     x11_start_program("firefox http://localhost", 60, { valid => 1 } );
     # starting from git might take a bit longer to get and generated assets
-    assert_screen 'openqa-dashboard', check_var('OPENQA_FROM_GIT', 1) ? 180 : 60;
+    # workaround for poo#19798, basically doubles the timeout
+    if ((check_screen 'openqa-dashboard', 180) == undef) {
+        record_soft_failure 'ff took to long to start';
+    }
+    #wait few minutes for ff to start and then fail the test
+    assert_screen 'openqa-dashboard', 360;
 }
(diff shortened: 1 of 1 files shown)


tests/openQA/dashboard.pm
```diff
@@ -9,7 +9,11 @@ sub run {
     ensure_unlocked_desktop();
     x11_start_program("firefox http://localhost", 60, { valid => 1 } );
     # starting from git might take a bit longer to get and generated assets
-    assert_screen 'openqa-dashboard', check_var('OPENQA_FROM_GIT', 1) ? 180 : 60;
+    # workaround for poo#19798, basically doubles the timeout
+    if ((check_screen 'openqa-dashboard', 180) == undef) {
```
okurz: hint for the future: Just take the return value of check_screen as boolean and don't compare for equality with undef, i.e.

```perl
if (!check_screen(…)) …
```

tests/openQA/dashboard.pm
```diff
@@ -9,7 +9,11 @@ sub run {
...
+        record_soft_failure 'ff took to long to start';
```
okurz: Don't record a soft failure without a bug reference. Either we accept something as given and implement the test accordingly or we report something as a bug, normally on bugzilla.opensuse.org and track the bugref with the record_soft_failure. Otherwise tests might turn yellow but no one will know about this.

BTW: I still think the issue reg. slow firefox start is worth to investigate more. Please ask openSUSE release managers or report a bug about this and try to check manually.

PR #28, opened 2017-07-10
Revert dashboard.pm to original state

DrMullings:
Improved and simplified code.
Readded dropped check for OPENQA_FROM_GIT to keep distinction

Verification run:
http://pinky.arch.suse.de/tests/53#

PR#27:
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/27

### DIFF
--- a/tests/openQA/dashboard.pm
+++ b/tests/openQA/dashboard.pm
@@ -9,11 +9,6 @@ sub run {
     ensure_unlocked_desktop();
     x11_start_program("firefox http://localhost", 60, { valid => 1 } );
     # starting from git might take a bit longer to get and generated assets
-    # workaround for poo#19798, basically doubles the timeout
-    if ((check_screen 'openqa-dashboard', 180) == undef) {
-        record_soft_failure 'ff took to long to start';
-    }
-    #wait few minutes for ff to start and then fail the test
-    assert_screen 'openqa-dashboard', 360;
+    assert_screen 'openqa-dashboard', check_var('OPENQA_FROM_GIT',1) ? 180 : 60;
 }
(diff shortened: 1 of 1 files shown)


okurz:
Please don't add a record_soft_failure without a bugref. Wait until you have one and add it then.

DrMullings:
Fixed Typo, still waiting for bug to be created so I can add it to the code

DrMullings:
new commit that contains the original state of dashboard.pm since the issue seems to be fixed

tests/openQA/dashboard.pm
```diff
@@ -9,11 +9,6 @@ sub run {
     ensure_unlocked_desktop();
     x11_start_program("firefox http://localhost", 60, { valid => 1 } );
     # starting from git might take a bit longer to get and generated assets
-    # workaround for poo#19798, basically doubles the timeout
-    if ((check_screen 'openqa-dashboard', 180) == undef) {
-        record_soft_failure 'ff took to long to start';
```
SergioAtSUSE: Please leave the record_soft_failure, it hasn't happened in this verification run, but it can happened again in the future. The only thing you have to add, it the bug reference (link to bugzilla).

We augmented the timeout to make the test green as workaround and complete other steps of the test. But that firefox take so much time to start is still for me a bug.

SergioAtSUSE: As spoken with @okurz, please ignore the above comment

DrMullings:
http://pinky.arch.suse.de/tests
Last 10 runs without problems, therefore I assume it is fixed and this revert can be merged

PR #32, opened 2017-08-18
Fix test_running by trying up to five times

DrMullings:
Fixed test running by looking up to five times with a waiting time
of 30 seconds between each try
Necessary due to slow starting test used in start_test

Verification runs:
[1](http://10.160.65.204/tests/494)
[2](http://10.160.65.204/tests/498)


DrMullings:
Further verification runs:
499-501

PR #33, opened 2017-12-07
Fix and simplify from_git routine:

coolo:
- install perl packages as rpms not from source
- install sass from rpm not from source
- leave out most devel packages
- start postgresql


coolo:
I forgot to mention: http://tortuga.suse.de/tests/20

PR #34, opened 2018-05-17
Delete obsolete vim footer in files

okurz:


kravciak:
obsolete? :)

okurz:
Not the right word? We deleted the footer in os-autoinst-distri-opensuse which I would consider a *reference* for os-autoinst-distri-openQA so if we do something there then we should do it here as well :)

PR #35, opened 2018-07-31
WIP: Handle grub without timeout

SergioAtSUSE:
openQA in openQA tests are failing because the grub menu has no timeout.

- https://openqa.opensuse.org/tests/overview?distri=openqa&version=Tumbleweed&build=%3ATW.1712&groupid=24

I am not able to get the tests properly configured to test locally.


okurz:
Regarding local tests probably you need to overwrite a test variable to use https://openqa.opensuse.org instead of http://openqa.opensuse.org

okurz:
Thank you for your PR but actually I think I have broken it with the recent deployment of the most recent os-autoinst version including setting a default check_screen timeout of 0 seconds. See in https://openqa.opensuse.org/tests/716490/file/autoinst-log.txt that the first check_screen now checks for (only) 0 seconds. A simple fix should be just to bring this back by setting explicit check_screen timeouts of 30 seconds on check_screen calls where we need unless you want to change it to an assert_screen with match_has_tag which should also work.

SergioAtSUSE:
@okurz, changed. But, I still have problems trying to do a local verification run...

PR #36, opened 2018-09-18
Add apache2 rewrite module

nicksinger:
Previously: https://openqa.opensuse.org/tests/757011#step/openqa_webui/29
Now: http://openqa.chell.qa.suse.de/tests/41#


okurz:
Awesome, thank you

PR #39, opened 2018-10-14
Add aarch64 support and fix test from git

ggardet:


ggardet:
@okurz this is the fixed version of https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/37

PR #42, opened 2019-05-14
Fix openqa_webui test

ggardet:
Verification run: https://openqa.opensuse.org/tests/933863

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -6,7 +6,7 @@ use testapi;
 sub get_log {
     my ($cmd, $name) = @_;
     my $ret = script_run "$cmd | tee $name";
-    upload_logs($name) if $ret;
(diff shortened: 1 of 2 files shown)


lib/openQAcoretest.pm
```diff
@@ -6,7 +6,7 @@ use testapi;
 sub get_log {
     my ($cmd, $name) = @_;
     my $ret = script_run "$cmd | tee $name";
-    upload_logs($name) if $ret;
+    upload_logs($name) if !$ret;
```
coolo: this would upload if it succeeded - I guess the point was indeed the opposite

ggardet: No, it must upload the log file only if it has been created successfully.
I tested it, and previously, the file were created, but not uploaded.

coolo:
I fetched manually on o3

PR #44, opened 2019-06-19
Fix tty switch; Refactor

okurz:
* Do not use perl method prototypes that should be avoided
* Delete unused `init_cmd` method
* Delete obsolete openSUSE < 42.3 support
* Rely on default 0 timeout in `check_screen`
* Save typing with `systemctl mask --now`
* Extract tty switch hotkeys
* Delete obsolete openSUSE 13.2 support
* Fix switch to text terminal in other modules as well

Verification run: https://openqa.opensuse.org/tests/963200


okurz:
Seems no one cares…

PR #45, opened 2019-06-28
Use standard "openqa-clone-job" instead of custom script

okurz:
Verification run: https://openqa.opensuse.org/t971381


okurz:
https://openqa.opensuse.org/t971381

PR #46, opened 2019-07-05
Fix test failure 53486

andrii-suse:
Fix https://progress.opensuse.org/issues/53486 according to changes in OpenQA PR#2110


okurz:
As this is a backward-compatible regex and you seem to have tested that already let's go ahead and merge it. Thanks!

PR #47, opened 2019-07-10
Use 'configure-web-proxy' instead of manual steps

okurz:
Verification runs:
* [openqa_install+publish](https://openqa.opensuse.org/tests/979134)
* [openqa_from_git](https://openqa.opensuse.org/tests/979161)


okurz:
For tests:
* https://openqa.opensuse.org/t979126
* https://openqa.opensuse.org/t979127

PR #53, opened 2019-11-19
Retry zypper installation

perlpunk:
Currently the installation can fail if the package is modified
during the command.

Issue: https://progress.opensuse.org/issues/60008
Test runs: https://openqa.opensuse.org/tests/1089700,  https://openqa.opensuse.org/tests/1089720, https://openqa.opensuse.org/tests/1089772

I verified that the exit code is 104 for `zypper install non-existant-package`. However, I'm not sure how to reproduce the actual error, so I don't know yet if the exit code will also be 104 there.

### DIFF
--- a/tests/install/test_distribution.pm
+++ b/tests/install/test_distribution.pm
@@ -9,7 +9,24 @@ sub run {
     diag('initialize working copy of openSUSE tests distribution with correct user');
     assert_script_run('username=bernhard email=bernhard@susetest /usr/share/openqa/script/fetchneedles', 3600);
     save_screenshot;
-    assert_script_run('zypper --no-refresh -n in os-autoinst-distri-opensuse-deps', 600);
+
+    # 104 is the zypper exit code if a package cannot be found
+    my $cmd = <<'EOM';
+(
+    for i in {1..7}
+    do
+        zypper --non-interactive install os-autoinst-distri-opensuse-deps && exit 0
+        rc=$?
+        (( $rc == 104 )) || break
+    done
+    exit $rc
+)
+EOM
+    chomp $cmd;
+    my $ret = assert_script_run(
+        $cmd,
+        timeout => 600,
+    );
     type_string "clear\n";
     # prepare for next test
     type_string "logout\n";


tests/install/test_distribution.pm
```diff
@@ -9,7 +9,24 @@ sub run {
     diag('initialize working copy of openSUSE tests distribution with correct user');
     assert_script_run('username=bernhard email=bernhard@susetest /usr/share/openqa/script/fetchneedles', 3600);
     save_screenshot;
-    assert_script_run('zypper --no-refresh -n in os-autoinst-distri-opensuse-deps', 600);
+
+    # 104 is the zypper exit code if a package cannot be found
+    my $cmd = <<'EOM';
+(
```
okurz: why the subshell?

tests/install/test_distribution.pm
```diff
@@ -9,7 +9,24 @@ sub run {
...
+    for i in {1..7}
+    do
```
okurz: ok to put the "do" on the same line as "for"?

tests/install/test_distribution.pm
```diff
@@ -9,7 +9,24 @@ sub run {
...
+        zypper --non-interactive install os-autoinst-distri-opensuse-deps && exit 0
+        rc=$?
+        (( $rc == 104 )) || break
```
okurz: how about `[ $sc = 104 ]`?

tests/install/test_distribution.pm
```diff
@@ -9,7 +9,24 @@ sub run {
...
+    done
+    exit $rc
```
okurz: Is this actually needed? If you just "break" from the loop the exit code should be preserved. The assignment `rc=$?` would break that. So would the following work?:

```sh
for i in {1..7}; do
    zypper --non-interactive install os-autoinst-distri-opensuse-deps && break || ( [ $? = 104 ] || break)
done
```

tests/install/test_distribution.pm
```diff
@@ -9,7 +9,24 @@ sub run {
...
+)
+EOM
+    chomp $cmd;
+    my $ret = assert_script_run(
+        $cmd,
+        timeout => 600,
+    );
```
okurz: A more common style – at least in our test code – is to write `assert_script_run $cmd, 600;`

And the variable assignment to $ret` is useless for two reasons:
1. You don't use `$ret`
2. `assert_script_run` will throw an exception if the assertion does not hold so use without return code evaluation.

tests/install/test_distribution.pm
perlpunk: `( [ $? = 104 ] || break)` does not wiork because the `break` is in the subshell

okurz:
Ah, right. Thx.

Am 20. November 2019 11:00:45 MEZ schrieb "Tina Müller (tinita)" <notifications@github.com>:

tests/install/test_distribution.pm
perlpunk: The documentation for `assert_script_run` is:
```
assert_script_run($cmd [, timeout => $timeout] [, fail_message => $fail_message] [,quiet => $quiet]);
```
AFAIK passing timeout as a second arg is using the old API which is still supported.

perlpunk: Using `$ret` was a leftover from experimenting, will remove it.

perlpunk:
I experimented with a little script to reproduce. The problem is that we need to have a comparison with `$?` and this will reset `$?`

We have three use cases.
* zypper exit code 0 should break the loop and succeed -> rc 0
* zypper exit code 104 should continue the loop -> rc != 0
* zypper exit code 1 (or other exit codes) should break and fail -> rc !=0

```
for i in 1 104 104 104; do
  echo "==== rc $i"
  perl -wE'$e = shift; exit $e' $i
  (( $? != 104 )) && break
done; echo "rc=$?"           
==== rc 1
rc=0         # <-- wrong
```
```
for i in 0 104 104 104; do
...
==== rc 0
rc=0      # <- ok
```
```
for i in 104 104 104; do
==== rc 104
==== rc 104
==== rc 104
rc=1      # <- wrong exit code, but ok for use case
```

As I didn't come up with code that does not reset `$?` I added the subshell.

tests/install/test_distribution.pm
perlpunk: Replied in a longer comment

perlpunk:
> Or else I would say the problem could happen the same on the other calls of zypper so then we should extract a method – like zypper_call in os-autoinst-distri-opensuse

I asked yesterday about using/copying methods from os-autoinst-distri-opensuse, but I was told to keep it simple and just use a bash loop ;-)

tests/install/test_distribution.pm
okurz: There are no plans to remove the positional argument but you are right, it says "Deprecated mode".

PR #54, opened 2019-11-20
Remove --no-refresh from zypper call

perlpunk:
Currently the install might fail if the package was updated in the meantime.

Issue: https://progress.opensuse.org/issues/60008
Test run: https://openqa.opensuse.org/tests/1090694

See also: https://github.com/openSUSE/zypper/issues/312


okurz:
Thx. We have yet to see if this is enough. Time will tell :)

PR #58, opened 2020-01-14
Force a zypper refresh before installing

perlpunk:
Issue: https://progress.opensuse.org/issues/62009

Test run: https://openqa.opensuse.org/tests/1143664

### DIFF
--- a/tests/install/openqa_worker.pm
+++ b/tests/install/openqa_worker.pm
@@ -5,8 +5,9 @@ use utils;
 
 sub run {
     diag('worker setup');
-    assert_script_run('zypper --no-cd --non-interactive --gpg-auto-import-keys in os-autoinst',   600);
-    assert_script_run('zypper --no-cd --non-interactive --gpg-auto-import-keys in openQA-worker', 600);
+    assert_script_run('zypper -n --gpg-auto-import-keys ref -f',           60);
+    assert_script_run('zypper --no-cd --non-interactive in os-autoinst',   600);
+    assert_script_run('zypper --no-cd --non-interactive in openQA-worker', 600);
     diag('Login once with fake authentication on openqa webUI to actually create preconfigured API keys for worker authentication');
     assert_script_run('curl http://localhost/login');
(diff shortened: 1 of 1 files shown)


lib/utils.pm
```diff
@@ -4,11 +4,12 @@ use base Exporter;
 use Exporter;
 use strict;
 use testapi;
+use File::Basename qw(basename);
 
 our @EXPORT = qw(switch_to_x11 wait_for_desktop ensure_unlocked_desktop);
 
 sub switch_to_x11 {
-    my @hdd = split(/-/, get_required_var('HDD_1'));
+    my @hdd = split(/-/, basename get_required_var('HDD_1'));
```
okurz: better if something like this is split into its own commit

tests/install/openqa_worker.pm
```diff
@@ -5,8 +5,9 @@ use utils;
 
 sub run {
     diag('worker setup');
-    assert_script_run('zypper --no-cd --non-interactive --gpg-auto-import-keys in os-autoinst',   600);
-    assert_script_run('zypper --no-cd --non-interactive --gpg-auto-import-keys in openQA-worker', 600);
+    assert_script_run('zypper -n --gpg-auto-import-keys ref -f',                                  60);
```
okurz: We don't have an automatic tidy check in this repo I assume but I doubt we need to keep the line like this. Looks weird if the timeout argument is separated by this many blanks. Also I suggest to just leave out the timeout and rely on the default which is 90s

perlpunk: Oops, that was a leftover. Fixed.

lib/utils.pm
perlpunk: Ok, I put it in its own branch

perlpunk:
Created #59 for the `basename`

lib/utils.pm
okurz: I am sorry if I did not make myself clear but I meant *just* **commit**, not its own PR or branch

PR #59, opened 2020-01-14
Get the basename of HDD before splitting

perlpunk:
A dash in the fullpath can get wrong results

Test run: https://openqa.opensuse.org/tests/1143729


okurz:
For the future, please see https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/58#discussion_r366439265 : I am sorry if I did not make myself clear but I meant just commit, not its own PR or branch

PR #62, opened 2021-03-11
Use validate_script_output to check the expected docker logs

Amrysliu:
When executing `system("$cmd logs $container 2>&1 | grep \"$text\" >/dev/null")`,
it will report `sh: /usr/bin/grep: Permission denied`.
see: https://openqa.opensuse.org/tests/1663505/file/autoinst-log.txt
This caused that we always waste 30s at that `while` loop.
We could simply call `validate_script_output` to do that.

Verify run: https://openqa.opensuse.org/tests/1663957#

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -87,18 +87,9 @@ sub ensure_unlocked_desktop {
 # - $container: The container name or ID
 # - $text: The text to search in the logs
 # - $cmd: The containers runner (docker, podman,...)
-# - $timeout: Time in seconds until this fails
 sub wait_for_container_log {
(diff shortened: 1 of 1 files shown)


okurz (commented):
Actually this looks like a good idea and also looks quite clean. @ilausuch as you were the author of the former code, your feedback would be appreciated as well

okurz (approved):
Actually this looks like a good idea and also looks quite clean. @ilausuch as you were the author of the former code, your feedback would be appreciated as well

Martchus (approved):

ilausuch (approved):

PR #63, opened 2021-03-12
containers: add more verbose when a container doesn't work well

ilausuch:
Problem: When a container doesn't work as expected and the function
wait_for_container_log fails with a timeout, we don't have logs to
analyze

Solution: On case of timeout the logs are displayed

Related to https://progress.opensuse.org/issues/88754

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -97,6 +97,8 @@ sub wait_for_container_log {
         last if ($timeout == 0);
     }
 
+    system("$cmd logs $container") if ($timeout == 0);
+
     assert_script_run("$cmd logs $container 2>&1 | grep \"$text\" >/dev/null");
 }
 


Martchus:
This change will conflict with #62. Note that @Amrysliu has requested your feedback on it and I'd also like to merge #62 first.

okurz:
Please

* Fix the conflict
* Show a verification run
* Ensure container tests are enabled again in production before applying more changes

ilausuch:
I think this won't be longer necessary since #62 is merged. This way to check the logs also print on the openqa job logs the problem if exists see https://openqa.opensuse.org/tests/1663957#step/worker/5

kalikiana:
> I think this won't be longer necessary since #62 is merged. This way to check the logs also print on the openqa job logs the problem if exists see https://openqa.opensuse.org/tests/1663957#step/worker/5

Are you going to close the PR then?

ilausuch:
> > I think this won't be longer necessary since #62 is merged. This way to check the logs also print on the openqa job logs the problem if exists see [openqa.opensuse.org/tests/1663957#step/worker/5](https://openqa.opensuse.org/tests/1663957#step/worker/5)
> 
> Are you going to close the PR then?

Yes closing

PR #67, opened 2021-04-21
containers: Show openqa_webui logs to ensure a correct investigation

ilausuch:
Openqa test "single_container_webui" eventually fails but we don't have information
about what happened because the logs are removed when the container is removed

Removing -rm ensures that we have the logs although it fails

https://progress.opensuse.org/issues/91488

### DIFF
--- a/tests/containers/single_container_webui.pm
+++ b/tests/containers/single_container_webui.pm
@@ -9,11 +9,16 @@ sub run {
 
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run --rm -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
   wait_for_container_log("openqa_webui", "Web application available at", "docker");
 
   assert_script_run("curl http://localhost");
   assert_script_run("docker rm -f openqa_webui");
+}
+
+sub post_fail_hook {
+  script_run("docker logs openqa_webui");
+  script_run("docker rm -f openqa_webui");
 }
 
 1;


tests/containers/single_container_webui.pm
```diff
@@ -9,9 +9,10 @@ sub run {
...
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run --rm -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
   wait_for_container_log("openqa_webui", "Web application available at", "docker");
-
+  assert_script_run("docker rm -f openqa_webui");
+  
   assert_script_run("curl http://localhost");
   assert_script_run("docker rm -f openqa_webui");
```
Martchus, line 17: This is now done twice. I'm not quite sure what the intention is but this looks quite odd.

ilausuch: Yes, was a mistake

ilausuch:
Now We have logs to research the problems in case this test fails
https://openqa.opensuse.org/tests/1708152#step/single_container_webui/7

tests/containers/single_container_webui.pm
```diff
@@ -9,7 +9,7 @@ sub run {
 
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run --rm -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
```
kalikiana, line 12: How does removing `--rm` give us logs? This *should* mean that the container isn't removed after the run, shouldn't it?

kalikiana: --rm                                       Remove container (and pod if created) after exit

ilausuch: You are right, it maintains the container created until it is destroyed manually using docker rm that happens in the last line of the run, or in the post_fail_hook if something is wrong

ilausuch: About the logs, I add a new line to print the logs in case of failure. I am not sure if show the logs if all are correct and pass the test

kalikiana: The *post_fail_hook* only applies if the test is considered failed, and I think that's the case here. If a container command fails the test also fails.

ilausuch:
I run this job https://openqa.opensuse.org/tests/1708943

kalikiana (approved):
We should probably double-check that this passes now. Although I almost want to see a failure now that we have the logging 😉

ilausuch:
> We should probably double-check that this passes now. Although I almost want to see a failure now that we have the logging wink

I launched this one https://openqa.opensuse.org/tests/1708948
Although I think we can force an error to test

Martchus (approved):

ilausuch:
I launched a run with intentional failure to see the results of this commit
https://openqa.opensuse.org/tests/1709015#step/single_container_webui/111

PR #68, opened 2021-04-23
Add search test written in Python

kalikiana:
See: https://progress.opensuse.org/issues/91257

Verification run: https://openqa.opensuse.org/t1840023

### DIFF
--- a/main.pm
+++ b/main.pm
@@ -59,6 +59,10 @@ sub load_openQA_tests() {
     }
 }
 
+sub load_python_tests() {
+    loadtest "openQA/search.py";
+}
+
 sub load_shutdown() {
     loadtest "shutdown/shutdown.pm";
 }
@@ -73,6 +77,7 @@ elsif (get_var('INSTALL')) {
 # testing from git only tests webui so far
 load_osautoinst_tests() unless check_var('OPENQA_FROM_GIT', 1);
 load_openQA_tests();
+load_python_tests() if get_var('PYTHON');
 load_shutdown();
 
 1;
--- a/tests/openQA/search.py
+++ b/tests/openQA/search.py
@@ -1,0 +1,22 @@
+from testapi import *
+
+
+def run(self):
+    assert_screen('openqa-logged-in')
+    assert_and_click('openqa-search')
+    type_string('shutdown.pm')
+    send_key('ret')
+    assert_screen('openqa-search-results')
+
+
+def switch_to_root_console():
+    send_key('ctrl-alt-f3')
+
+
+def post_fail_hook(self):
+    switch_to_root_console()
+    assert_script_run('openqa-cli api experimental/search q=shutdown.pm')
+
+
+def test_flags(self):
+    return {'fatal': 1}


grisu48 (commented):
This looks already very promising!

okurz:
actually good idea :)

tests/openQA/search.py
```diff
@@ -0,0 +1,34 @@
...
+    assert_screen('openqa-logged-in')
+    assert_and_click('openqa-search')
+    type_string('shutdown.pm')
+    send_key('ret')
+    assert_screen('openqa-search-results')
+
+
+def post_fail_hook(self):
+    send_key('ctrl-alt-f3')  # switch to root console
+    script_run('openqa-cli api --o3 experimental/search q=shutdown.pm')
```
okurz, line 29: as you run within the openQA-in-openQA instance that setup an openQA instance I recommend to connect to that, not "o3". As "localhost" is default I suggest to simply delete the parameter and then have a test that works with that instance.

tests/openQA/search.py
```diff
@@ -0,0 +1,34 @@
...
+    save_screenshot()
```
okurz, line 30: Isn't there a screenshot already after the script_run call finished?

tests/openQA/search.py
```diff
@@ -0,0 +1,33 @@
...
+    assert_screen('openqa-search-results')
+
+
+def post_fail_hook(self):
+    send_key('ctrl-alt-f3')  # switch to root console
+    script_run('openqa-cli api experimental/search q=shutdown.pm')
+
+
+def test_flags(self):
+    return dict([('fatal', 1)])
```
Martchus, line 33:
```suggestion
    return {'fatal': 1}
```

main.pm
```diff
@@ -59,6 +59,11 @@ sub load_openQA_tests() {
     }
 }
 
+sub load_python_tests() {
+    # use Inline Python => "import os.path, sys; sys.path.insert(0, os.path.abspath(os.path.join(os.path.curdir, '../..')))";
```
okurz, line 63: delete dead code

tests/openQA/search.py
```diff
@@ -0,0 +1,33 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, see <http://www.gnu.org/licenses/>.
+
```
okurz, line 15: Why do you want to introduce a GPL licensed file? Other files within the repo don't have that license header and the project is MIT-licensed. I suggest to simply delete the header

tests/openQA/search.py
```diff
@@ -0,0 +1,33 @@
...
+def run(self):
+    assert_screen('openqa-logged-in')
+    assert_and_click('openqa-search')
+    type_string('shutdown.pm')
+    send_key('ret')
+    assert_screen('openqa-search-results')
+
+
+def post_fail_hook(self):
+    send_key('ctrl-alt-f3')  # switch to root console
```
okurz, line 28:
```suggestion

def switch_to_root_console():
    send_key('ctrl-alt-f3')

def post_fail_hook(self):
```

tests/openQA/search.py
```diff
@@ -0,0 +1,33 @@
...
+    script_run('openqa-cli api experimental/search q=shutdown.pm')
```
okurz, line 29: Please use `assert_script_run` unless you want to check the exit code explicitly

okurz (changes requested):

okurz (approved):
This looks great now and could act as a very good example!

Care to mention it in the openQA documentation as showcase example? :)

PR #71, opened 2021-07-14
[WIP] containers: pre-create the working directories with a=rwx permissions 

ilausuch:
We found that eventually the test single_container_webui fails because the webui
cannot create the needed directories in /root/data/factory

With this commit we ensure to create the directories with the correct permissions
before launch the container

Testing: https://openqa.opensuse.org/tests/1840015

https://progress.opensuse.org/issues/95179

### DIFF
--- a/tests/containers/single_container_webui.pm
+++ b/tests/containers/single_container_webui.pm
@@ -6,7 +6,8 @@ sub run {
   my ($self) = @_;
   my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/webui/conf:/data/conf:ro"';
   my $certificates = '-v "/root/server.crt:/etc/apache2/ssl.crt/server.crt" -v "/root/server.crt:/etc/apache2/ssl.crt/ca.crt" -v "/root/server.key:/etc/apache2/ssl.key/server.key"';
-
+  
+  assert_script_run("mkdir -p -m a=rwx /root/data/{factory,tests}");
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
   assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");


Martchus (changes requested):
I've mentioned in https://progress.opensuse.org/issues/95179#note-7 that the test doesn't fail because of the permission errors. Fixing the permission error makes of course nevertheless sense - but don't state the test failure is fixed by that. (The permission error also happens in successful tests.) You can still reference the ticket in the commit message, but reference it like "The permission errors have been observed when investigating https://progress.opensuse.org/issues/95179.".

tests/containers/single_container_webui.pm
```diff
@@ -4,9 +4,10 @@ use utils;
 
 sub run {
   my ($self) = @_;
-  my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/webui/conf:/data/conf:ro"';
+  my $volumes = '"-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/webui/conf:/data/conf:ro""';
```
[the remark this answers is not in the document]
ilausuch: Yes, and the test didn't failed

ilausuch:
Yes, but I mentioned that it could be interesting to have this. In local tests without the correct permissions is problematic to me

ilausuch:
I defend that we should consider this PR, because this problem is not affecting now the test but could have negative effects in futures uses of this test

tests/containers/single_container_webui.pm
```diff
@@ -6,7 +6,8 @@ sub run {
   my ($self) = @_;
   my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/webui/conf:/data/conf:ro"';
   my $certificates = '-v "/root/server.crt:/etc/apache2/ssl.crt/server.crt" -v "/root/server.crt:/etc/apache2/ssl.crt/ca.crt" -v "/root/server.key:/etc/apache2/ssl.key/server.key"';
-
+  
+  assert_script_run("mkdir -p -m a=rwx /root/data/{factory,tests}");
```
okurz, line 10: hm, do we really only need this in tests or do users run into this as well?

Martchus:
> I defend that we should consider this PR

You don't need to. I said "Fixing the permission error makes of course nevertheless sense". I was only saying that the commit message should be changed: "… don't state the test failure is fixed by that" and "You can still reference the ticket in the commit message, but reference it like "The permission errors have been observed when investigating https://progress.opensuse.org/issues/95179.".".

bmwiedemann:
In general `777` permissions are a red flag for you doing unsafe things. Even `/tmp` has `1777` to get more security.

You should just give the permission that is needed there, e.g. allow one other user access there: `setfacl -m u:USERNAME:rwx /root/data/factory`

tests/containers/single_container_webui.pm
ilausuch: If they run this using directly without the docker-compose yes, users will have the same problem because the openQA is going to be executed using the geekotest user and the proprietary of these directories inside the container is root

I did a test creating the tests and factory directories as a non-root user
```
drwxr-xr-x  2 suse      suse    64B Jul 19 14:01 factory
drwxr-xr-x  2 suse      suse    64B Jul 19 14:01 tests
```

But inside the container the owner is root, so yes, will be a problem for the user too

```
144255a00bf5:/data # ls -lh
total 0
drwxr-xr-x 2 root root 40 Jul 19 11:57 factory
drwxr-xr-x 2 root root 40 Jul 19 11:57 tests
```

ilausuch:
I'm testing @bmwiedemann idea

tests/containers/single_container_webui.pm
okurz: My point is: If it's an issue that users could also hit then don't just apply a fix in tests but in the underlying project, i.e. openQA itself

ilausuch:
I think a best option is this https://github.com/os-autoinst/openQA/pull/4056

ilausuch:
closing this ticket, continues on os-autoinst/openQA#4056

PR #73, opened 2021-08-04
[WIP] Replace send_key by send_key_until_needlematch to change to the console

ilausuch:
Sometimes send_key is not enough to force the change to the console failing
the console needle check.

https://progress.opensuse.org/issues/95721

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -75,7 +75,7 @@ sub install_containers {
 
 sub run {
     send_key "ctrl-alt-f3";
-    assert_screen "inst-console";
+    assert_screen("inst-console", timeout => 60);
     type_string "root\n";
     assert_screen "password-prompt";
     type_string $testapi::password . "\n";


ilausuch:
Test https://openqa.opensuse.org/t1867956

okurz (changes requested):
Have you seen https://progress.opensuse.org/issues/95721#note-3 ? This mentioned that the hotkey was effective but that the test did not wait long enough. You are doing something quite different. Also in https://progress.opensuse.org/issues/95721#note-10 you provided an example of `send_key 'ret'` which is again something different

tests/install/openqa_webui.pm
```diff
@@ -74,8 +74,9 @@ sub install_containers {
 }
 
 sub run {
-    send_key "ctrl-alt-f3";
-    assert_screen "inst-console";
+    #send_key "ctrl-alt-f3";
+    #assert_screen "inst-console";
+    send_key_until_needlematch("inst-console", "ctrl-alt-f3", 5, 1);
```
kalikiana, line 79: I would expect [send_key_until_needlematch](http://open.qa/api/testapi/#_send_key_until_needlematch) to hit they key up to 5 times with a 1 second delay.
Didn't the failure show that the console got opened? So we might just need a >30s timeout for `assert_screen`?

ilausuch: Yes, but only if send_key works perfectly. Anyway, I change that assuming that the send_key is not the problem

ilausuch: I changed that and I created this test https://openqa.opensuse.org/tests/1868785

kalikiana: You changed it to `timeout => 30` - that is the default, so it would be exactly the same :-D

okurz:
this isn't progressing and I guess @ilausuch has no plans to continue

PR #75, opened 2021-11-18
Also check for 'done' jobs

perlpunk:
This has been failing a lot recently (returning an empty list), and it seems
the jobs are getting done too fast to check for a running state.

https://openqa.opensuse.org/group_overview/24?limit_builds=20&limit_builds=50

### DIFF
--- a/tests/osautoinst/test_running.pm
+++ b/tests/osautoinst/test_running.pm
@@ -4,7 +4,7 @@ use testapi;
 
 sub run {
     assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
-    assert_script_run 'ret=false; for i in {1..5} ; do openqa-client jobs state=running | ack --passthru --color running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
+    assert_script_run 'ret=false; for i in {1..5} ; do openqa-cli api jobs state=running state=done | ack --passthru --color 'running|done' && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
     save_screenshot;
     type_string "clear\n";
 }


perlpunk:
Looking at what is displayed after the failing step suggests that the problem is not that it takes too long until the job is running, but it is already done at this point:
https://openqa.opensuse.org/tests/2043020#step/test_running/7

perlpunk:
That's what I was trying to say with my last comment?

Martchus:
Yes, I understand that. I was just stating the consequence of your comment explicitly.

tests/osautoinst/test_running.pm
```diff
@@ -4,7 +4,7 @@ use testapi;
 
 sub run {
     assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
-    assert_script_run 'ret=false; for i in {1..5} ; do openqa-client jobs state=running | ack --passthru --color running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
+    assert_script_run 'ret=false; for i in {1..10} ; do openqa-client jobs state=running | ack --passthru --color running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
```
okurz, line 7: 10*30 plus overhead exceeds the timeout of 300s. Better increase that as well then if you really want to bump the retries

perlpunk: No I don't want to, that's why I converted it to a **draft**

perlpunk:
I changed the `ack` to look for running and done

tests/osautoinst/test_running.pm
```diff
@@ -4,7 +4,7 @@ use testapi;
 
 sub run {
     assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
-    assert_script_run 'ret=false; for i in {1..5} ; do openqa-client jobs state=running | ack --passthru --color running && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
+    assert_script_run 'ret=false; for i in {1..5} ; do openqa-client jobs state=running | ack --passthru --color 'running|done' && ret=true && break ; sleep 30 ; done ; [ "$ret" = "true" ]', 300;
```
okurz, line 7: if this works, does that mean that `openqa-client jobs state=running` does not really filter for "running" and instead simply shows all jobs?

perlpunk: oops. , how embarassing.

I now added `state=done` and also switched to the new client.

Martchus (approved):

okurz (approved):

okurz:
https://openqa.opensuse.org/tests/2057593 looks related:

```
Bareword found where operator expected at openqa//tests/osautoinst/test_running.pm line 7, near "'ret=false; for i in {1..5} ; do openqa-cli api jobs state=running state=done | ack --passthru --color 'running"
	(Missing operator before running?)
[2021-11-26T07:25:34.610213Z] [warn] !!! autotest::loadtest: error on /tests/osautoinst/test_running.pm: Bad name after done' at openqa//tests/osautoinst/test_running.pm line 7.
  Compilation failed in require at (eval 125) line 1.
  
error on /tests/osautoinst/test_running.pm: Bad name after done' at openqa//tests/osautoinst/test_running.pm line 7.

```

Martchus:
Yes, the string literal quoting is broken.

PR #79, opened 2022-02-10
Shutdown stability

mimi1vx:
https://progress.opensuse.org/issues/105506

Added clear command to tests using same console as shutdown to make needle match more stable

https://openqa.opensuse.org/tests/2181124#step/shutdown/1

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -6,7 +6,12 @@ use strict;
 use testapi;
 use File::Basename qw(basename);
 
-our @EXPORT = qw(switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log);
+our @EXPORT = qw(clear_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log);
+
+sub clear_root_console {
+    enter_cmd('clear');
+    assert_screen 'root-console';
+}
 
 sub switch_to_x11 {
     my @hdd = split(/-/, basename get_required_var('HDD_1'));
--- a/tests/install/test_distribution.pm
+++ b/tests/install/test_distribution.pm
@@ -11,9 +11,9 @@ sub run {
     save_screenshot;
     assert_script_run('zypper -n ref -f',                              60);
     assert_script_run('zypper -n in os-autoinst-distri-opensuse-deps', 600);
-    type_string "clear\n";
+    clear_root_console;
     # prepare for next test
-    type_string "logout\n";
+    enter_cmd "logout";
     switch_to_x11;
 }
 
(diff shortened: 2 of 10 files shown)


tests/shutdown/shutdown.pm
```diff
@@ -3,8 +3,9 @@ use base "openQAcoretest";
 use testapi;
 
 sub run {
-    wait_screen_change { send_key 'ctrl-alt-f3' };
-    type_string "poweroff\n";
+    send_key('ctrl-alt-f3', wait_scren_change => 1) ;
```
okurz, line 6: that still has a typo.

tests/shutdown/shutdown.pm
```diff
@@ -3,8 +3,9 @@ use base "openQAcoretest";
 use testapi;
 
 sub run {
-    wait_screen_change { send_key 'ctrl-alt-f3' };
-    type_string "poweroff\n";
+    send_key('ctrl-alt-f3', wait_screen_change => 1) ;
```
Martchus, line 6: Is `wait_screen_change => 1` required if we're using an `assert_screen` in the next line anyways?

okurz: No, shouldn't be required. That was my point

lib/utils.pm
```diff
@@ -6,7 +6,12 @@ use strict;
 use testapi;
 use File::Basename qw(basename);
 
-our @EXPORT = qw(switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log);
+our @EXPORT = qw(clear_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log);
+
+sub clear_root_console {
+    enter_cmd("clear");
+    assert_screen "root-console";
```
okurz, line 13: please prefer single tickets as long as you don't want perl code to be evaluated. So better

```suggestion
    enter_cmd('clear');
    assert_screen 'root-console';
```

tests/install/test_distribution.pm
```diff
@@ -11,9 +11,10 @@ sub run {
     save_screenshot;
     assert_script_run('zypper -n ref -f',                              60);
     assert_script_run('zypper -n in os-autoinst-distri-opensuse-deps', 600);
-    type_string "clear\n";
+    enter_cmd("clear");
+    assert_screen "root-console";
```
okurz, line 15:
```suggestion
    clear_root_console;
```

okurz (changes requested):

Martchus (approved):

lib/utils.pm
okurz: seems you overlooked some other changes where there are still double ticks. Please check again (or I did something wrong?)

okurz (approved):
You still did introduce double quotes for strings that you don't want to expand variables in. Apparently I can not make myself clear. So fine, let's merge it anyways :)

PR #80, opened 2022-02-16
Extend timeout for ISO download and show progress

Martchus:
The test recently ran into a timeout here:
```
Test died: command 'openqa-clone-job --from http://openqa.opensuse.org $job_id' timed out at openqa//tests/osautoinst/start_test.pm line 23.
```

Likely it just too longer than 90 seconds to download the ISO file.

See https://progress.opensuse.org/issues/106919


Martchus:
Verification run: https://w3.suse.de/~mkittler/job-2359.webm

Of course this is more a speed test for my internet connection. However, it shows that `--show-progress` doesn't disturb anything.

kalikiana (approved):

PR #81, opened 2022-03-23
Allow more time for sporadic zypper and download slowdown

okurz:
In multiple occassions we found zypper calls to take a very long time in
both the webui as well as worker preparation steps which involve
downloading multiple packages. This commit bumps the timeout from 1000s
to a full hour to account for random slowness.

Related progress issue: https://progress.opensuse.org/issues/108665


Martchus (approved):
So we might wait for three hours to build these containers. That seems horribly slow but I guess we have to accept it.

PR #84, opened 2022-05-11
Retry setting up the postgres container

kalikiana:
Fixes: https://progress.opensuse.org/issues/110803

### DIFF
--- a/tests/containers/setup_env.pm
+++ b/tests/containers/setup_env.pm
@@ -7,7 +7,7 @@ sub run {
 
    assert_script_run("mkdir -p /root/data/factory/{iso,hdd,other} /root/data/tests");
    assert_script_run("docker network create testing");
-   assert_script_run("docker run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name db postgres", timeout => 600);
+   assert_script_run("for i in {1..3}; do docker run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name db postgres && break; done", timeout => 600);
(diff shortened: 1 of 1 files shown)


kalikiana:
> Did you do a verification run?

Yes. But apparently `install_deps` is never run: https://openqa.opensuse.org/tests/2339622#step/setup_env/7

kalikiana:
> it probably needs to be installed the same way as docker in tests/install/openqa_webui.pm
> 
> ```
> assert_script_run("zypper --non-interactive install docker git", timeout => 600);
> ```

That makes a lot of sense in hindsight since it's run as part of the test 🤦🏾 Thanks

tests/install/openqa_webui.pm
```diff
@@ -53,7 +53,7 @@ EOF
 }
 
 sub install_containers {
-    assert_script_run("zypper --non-interactive install docker git", timeout => 600);
+    assert_script_run("zypper --non-interactive install docker git podman", timeout => 600);
```
Martchus, line 56: Have you checked whether we still use docker elsewhere? If no, I suppose docker can be removed from here.

kalikiana:
> That makes a lot of sense in hindsight since it's run as part of the test 🤦🏾 Thanks

Successful verification run: https://openqa.opensuse.org/tests/2339705

tests/install/openqa_webui.pm
kalikiana: Yes, and I'm proposing #85 separately.

perlpunk (approved):

okurz:
This looks fine and I prefer it over #86 which can't work within os-autoinst-distri-openQA anyway.

kalikiana:
> This looks fine and I prefer it over #86 which can't work within os-autoinst-distri-openQA anyway.

It does work. Please keep an open mind when I am trying something out which takes about as much time as the whole conversation 😃 

Anyway fine by me to go with this version.

okurz:
Yes, you fixed it by including a valid definition of `script_retry` now

PR #85, opened 2022-05-11
Use podman everywhere

kalikiana:


tests/containers/setup_env.pm
```diff
@@ -6,9 +6,9 @@ sub run {
    my ($self) = @_;
 
    assert_script_run("mkdir -p /root/data/factory/{iso,hdd,other} /root/data/tests");
-   assert_script_run("docker network create testing");
-   assert_script_run("docker run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name db postgres", timeout => 600);
-   wait_for_container_log("db", "database system is ready to accept connections", "docker");
+   assert_script_run("podman network create testing");
+   assert_script_run("for i in {1..3}; do podman run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --network-alias=db --name db docker.io/library/postgres:latest && break; done", timeout => 600);
```
grisu48, line 10: Just as a suggestion: why not use `script_retry` with a `delay => 60` here? A delay between retries has been proven useful, especially when dealing with network-related issues where immediate retries often fail in the same way.

```suggestion
   script_retry("podman run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --network-alias=db --name db docker.io/library/postgres:latest", retry=> 3, delay => 60, timeout => 600);
```

okurz: Oh, this is where sneaky @grisu48 led you into a trap :) So you tried that in https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/86 but it can't directly work because `script_retry` is only defined within the scope of os-autoinst-distri-opensuse != os-autoinst-distri-openQA

okurz (changes requested):
I like the change but I would like to see a proof that this works.

kalikiana:
> I like the change but I would like to see a proof that this works.

It works, but I wonder why [podman build times out](https://openqa.opensuse.org/tests/2353545#step/build/3) so easily. Is this normal and just hidden behind retries? There's a timeout of 3600 and it seems to simply take longer getting all of the packages installed.

tests/containers/setup_env.pm
```diff
@@ -6,9 +6,9 @@ sub run {
    my ($self) = @_;
 
    assert_script_run("mkdir -p /root/data/factory/{iso,hdd,other} /root/data/tests");
-   assert_script_run("docker network create testing");
-   assert_script_run("for i in {1..3}; do docker run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --net-alias=db --name db postgres && break; done", timeout => 600);
-   wait_for_container_log("db", "database system is ready to accept connections", "docker");
+   assert_script_run($self->{cre} . " network create testing");
+   assert_script_run($self->{cre} . " run --rm -d --network testing -e POSTGRES_PASSWORD=openqa -e POSTGRES_USER=openqa -e POSTGRES_DB=openqa --network-alias=db --name db docker.io/library/postgres:latest", timeout => 600);
```
okurz, line 10: seems you are reverting the retrying you added just recently?

kalikiana: I can't see it in the diff, but likely this was a timeout I mistakenly added when I didn't know where it was timing out

okurz: But … you removed the retrying, see the diff please. The `for i in {1..3}; … ` is gone

okurz (changes requested):

PR #87, opened 2022-05-26
Add AutoConfig for Firefox to disable pop-ups on start

baierjan:
Reference: https://progress.opensuse.org/issues/111545

Did a few verification runs on http://polaris.suse.cz/tests/2506#next_previous, no pop-up since I added the configuration.

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -6,7 +6,7 @@ use strict;
 use testapi;
 use File::Basename qw(basename);
 
-our @EXPORT = qw(clear_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log);
+our @EXPORT = qw(clear_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log prepare_firefox_autoconfig);
 
 sub clear_root_console {
     enter_cmd('clear');
@@ -104,6 +104,26 @@ sub wait_for_container_log {
         sleep 1;
     }
     validate_script_output("$cmd logs $container 2>&1", qr/$text/);
+}
+
+# Use AutoConfig file for firefox to predefine some user values
+# https://support.mozilla.org/en-US/kb/customizing-firefox-using-autoconfig
+sub prepare_firefox_autoconfig {
+    # Enable AutoConfig by pointing to a cfg file
+    type_string(q{cat <<EOF > $(rpm --eval %_libdir)/firefox/defaults/pref/autoconfig.js
+pref("general.config.filename", "firefox.cfg");
+pref("general.config.obscure_value", 0);
+EOF
+});
+    # Create AutoConfig cfg file
+    type_string(q{cat <<EOF > $(rpm --eval %_libdir)/firefox/firefox.cfg
+// Mandatory comment
+// https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/first-run.html
+pref("browser.aboutwelcome.enabled", false);
+pref("browser.startup.upgradeDialog.enabled", false);
+pref("privacy.restrict3rdpartystorage.rollout.enabledByDefault", false);
+EOF
+});
 }
 
 1;
--- a/tests/openQA/dashboard.pm
+++ b/tests/openQA/dashboard.pm
@@ -4,6 +4,7 @@ use testapi;
 use utils;
 
 sub run {
+    prepare_firefox_autoconfig;
     switch_to_x11;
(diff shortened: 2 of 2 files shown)


DimStar77:
Can this also be used to address https://progress.opensuse.org/issues/111416?

baierjan:
Yes, I believe the same approach can be used also for os-autoinst-distri-opensuse. The biggest challenge would be to find a proper place where to include the creation of the config files.

Martchus (approved):

perlpunk (approved):

lib/utils.pm
```diff
@@ -106,4 +106,18 @@ sub wait_for_container_log {
...
+sub prepare_firefox_autoconfig {
+    # Enable AutoConfig by pointing to a cfg file
+    assert_script_run(q{echo 'pref("general.config.filename", "firefox.cfg");' > $(rpm --eval %_libdir)/firefox/defaults/pref/autoconfig.js});
+    assert_script_run(q{echo 'pref("general.config.obscure_value", 0);' >> $(rpm --eval %_libdir)/firefox/defaults/pref/autoconfig.js});
+    # Create AutoConfig cfg file
+    assert_script_run(q{echo '// Mandatory comment' > $(rpm --eval %_libdir)/firefox/firefox.cfg});
+    # https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/first-run.html
+    assert_script_run(q{echo 'pref("browser.aboutwelcome.enabled", false);' >> $(rpm --eval %_libdir)/firefox/firefox.cfg});
+    assert_script_run(q{echo 'pref("browser.startup.upgradeDialog.enabled", false);' >> $(rpm --eval %_libdir)/firefox/firefox.cfg});
+    assert_script_run(q{echo 'pref("privacy.restrict3rdpartystorage.rollout.enabledByDefault", false);' >> $(rpm --eval %_libdir)/firefox/firefox.cfg});
```
okurz, line 120: I suggest to use a here doc so that there is only one command for each file

baierjan: That is a good idea! Did you mean bash here doc or a perl one?

The problem I see is that `assert_script_run` add the fancy `; echo $?` at the end and that collides with the end of the here doc. Any suggestions around this? I can probably swap `script_run` with `type_string` or `enter_cmd`, but then I will lose the assert.

okurz: well, I am thinking of a perl here-doc with an echo call in bash using control characters similar as in https://github.com/os-autoinst/os-autoinst-distri-openQA/blob/master/tests/install/openqa_webui.pm#L20 :

```perl
# https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/first-run.html
my $firefox_config = <<'EOF';
echo -e "pref(\"general.config.filename\", \"firefox.cfg\");\npref(\"general.config.obscure_value\", 0);" > $(rpm --eval %_libdir)/firefox/defaults/pref/autoconfig.js
echo -e "// Mandatory comment\npref("browser.aboutwelcome.enabled", false);\npref("browser.startup.upgradeDialog.enabled", false);\npref("privacy.restrict3rdpartystorage.rollout.enabledByDefault", false);" > $(rpm --eval %_libdir)/firefox/firefox.cfg
EOF
    assert_script_run($_) foreach (split /\n/, $firefox_config);
```

granted. It would look cleaner if we could also use a bash here-doc to separate the lines that we want to write to each script.

baierjan: Yeah, I saw the perl here doc we have, but unless I create long echo line, I still need to invoke multiple commands.

Maybe I can offer this:
```perl
sub prepare_firefox_autoconfig {
    # Enable AutoConfig by pointing to a cfg file
    type_string(q{cat <<EOF > $(rpm --eval %_libdir)/firefox/defaults/pref/autoconfig.js
pref("general.config.filename", "firefox.cfg");
pref("general.config.obscure_value", 0);
EOF
});
    # Create AutoConfig cfg file
    type_string(q{cat <<EOF > $(rpm --eval %_libdir)/firefox/firefox.cfg
// Mandatory comment
// https://firefox-source-docs.mozilla.org/browser/components/newtab/content-src/asrouter/docs/first-run.html
pref("browser.aboutwelcome.enabled", false);
pref("browser.startup.upgradeDialog.enabled", false);
pref("privacy.restrict3rdpartystorage.rollout.enabledByDefault", false);
EOF
});
}
```

okurz: Looks good. We can actually also upload files but your approach should be good

okurz (approved):

PR #96, opened 2022-08-03
Drop x11_start_program from susedistribution

kalikiana:
See: https://progress.opensuse.org/issues/108530


okurz (approved):
As I confirmed that the string "x11_start_program" does not appear anywhere else in this test distribution I can approve without verification run

perlpunk:
the `use testapi ...` line is obsolete now too, I guess :)

PR #99, opened 2022-09-01
Improve error handling in container build tests

Martchus:
The for-loop on Bash level would not make the execution stop
when the webui-container build fails. That makes it hard to trace
what the source of the problem was.

This doesn't fix the root cause of
https://progress.opensuse.org/issues/116107 but should make
debugging easier.


perlpunk (approved):

Martchus:
Verification run: https://openqa.opensuse.org/tests/2598364

perlpunk:
the first VR failed because of an unrelated failure. retry: https://openqa.opensuse.org/tests/2598403

kalikiana (approved):

PR #100, opened 2022-10-06
Ensure repositories are refreshed why trying to install openQA packages

Martchus:
This should fix failures when the openQA package has been updated since the last refresh. Otherwise the retry just runs into the same issue again as by default the automatic refresh is only triggered every 10 minutes (but we try only 3 times with a delay of 30 seconds).

See https://progress.opensuse.org/issues/117280


perlpunk (approved):

tests/install/openqa_webui.pm
```diff
@@ -16,7 +16,7 @@ sub install_from_repos {
     my $repo = 'openSUSE_' . $repo_suffix{get_required_var('ARCH')};
     $add_repo = "zypper -n ar -p 95 -f obs://devel:openQA/$repo openQA";
     assert_script_run($_) foreach (split /\n/, $add_repo);
-    assert_script_run('retry -s 30 -- zypper --no-cd -n --gpg-auto-import-keys in openQA-local-db', 600);
+    assert_script_run('retry -s 30 -- zypper -n ref && zypper --no-cd -n --gpg-auto-import-keys in openQA-local-db', 600);
```
okurz, line 19:
```suggestion
    assert_script_run('retry -s 30 -- sh -c "zypper -n ref && zypper --no-cd -n --gpg-auto-import-keys in openQA-local-db"', 600);
```

same below

okurz (changes requested):

tests/install/openqa_webui.pm
okurz: same as already done in tests/install/test_distribution.pm

okurz (approved):

PR #104, opened 2022-12-06
Avoid openQA worker from appearing broken due to no FQDN

Martchus:
See https://progress.opensuse.org/issues/121567

---

Still a draft because I need to do a verification run (and my local attempts so far were not successful as the test even fails before).

### DIFF
--- a/tests/install/openqa_worker.pm
+++ b/tests/install/openqa_worker.pm
@@ -8,7 +8,7 @@ sub run {
     assert_script_run('retry -s 30 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 600);
     diag('Login once with fake authentication on openqa webUI to actually create preconfigured API keys for worker authentication');
     assert_script_run('curl http://localhost/login');
-    diag('adding temporary, preconfigured API keys to worker config');
+    diag('adding temporary, preconfigured API keys to client config');
     type_string('cat >> /etc/openqa/client.conf <<EOF
 [localhost]
 key = 1234567890ABCDEF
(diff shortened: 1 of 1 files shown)


tests/install/openqa_worker.pm
```diff
@@ -17,6 +17,7 @@ EOF
 ');
     wait_still_screen(1);
     my $worker_setup = <<'EOF';
+echo "WORKER_HOSTNAME=127.0.0.1" >> /etc/openqa/workers.ini
```
okurz, line 20: but why do we need this? Shouldn't every sane openSUSE system be able to resolve "localhost" to 127.0.0.1?

perlpunk (approved):

tests/install/openqa_worker.pm
okurz: ok, `localhost` is nicer but I want to consider this in a bit more detail as the openQA-in-openQA tests can also be treated as instructions how systems are setup. What does it mean that we need to add this here? Who else under which circumstances would need this? Does everyone setting up openQA on a virtual machine would suddenly need this as well?

okurz (changes requested):
see https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/104?notification_referrer_id=NT_kwDOABnW-LI1MDEzMzMyNTUyOjE2OTM0MzI&notifications_query=is%3Aunread#discussion_r1041172924

tests/install/openqa_worker.pm
Martchus: That's how 2. would look like: https://github.com/os-autoinst/openQA/pull/4945

okurz: >     2. Enable the feature only when `WORKER_HOSTNAME = auto` is present in the worker config. It is the same as 1. but we'd avoid an additional setting. Of course it is problematic when the value should actually be "auto".

Do you mean it prevents us to come up with an actual hostname "auto"? I think I can live with that restriction, my host is called `true; DROP TABLE passwords` ;)

Martchus:
https://github.com/os-autoinst/openQA/pull/4949 is a better alternative

PR #106, opened 2023-01-23
Parse docker build output and look for known error

baierjan:
Reference: https://progress.opensuse.org/issues/123106

Failed job example: http://polaris.suse.cz/tests/21

### DIFF
--- a/tests/containers/build.pm
+++ b/tests/containers/build.pm
@@ -5,8 +5,8 @@ use testapi;
 sub run {
     my ($self) = @_;
     assert_script_run("git clone https://github.com/os-autoinst/openQA.git", timeout => 300);
-    assert_script_run("retry -s 30 -e -- docker build openQA/container/$_ -t openqa_$_", timeout => 2400) for qw(webui worker);
-    assert_script_run('retry -s 30 -e -- docker build openQA/container/openqa_data -t openqa_data', timeout => 3600);
+    assert_script_run("retry -s 30 -r 7 -e -- docker build openQA/container/$_ -t openqa_$_", timeout => 3800) for qw(webui worker);
+    assert_script_run('retry -s 30 -r 7 -e -- docker build openQA/container/openqa_data -t openqa_data', timeout => 3800);
 }
 
 1;


tests/containers/build.pm
```diff
@@ -5,8 +5,13 @@ use testapi;
 sub run {
     my ($self) = @_;
     assert_script_run("git clone https://github.com/os-autoinst/openQA.git", timeout => 300);
-    assert_script_run("retry -s 30 -e -- docker build openQA/container/$_ -t openqa_$_", timeout => 2400) for qw(webui worker);
-    assert_script_run('retry -s 30 -e -- docker build openQA/container/openqa_data -t openqa_data', timeout => 3600);
+    for (qw(webui worker)) {
+        my $output = script_output("retry -s 60 -e -- docker build openQA/container/$_ -t openqa_$_ && echo EXIT_CODE:\$?", 2400, proceed_on_failure => 1);
```
okurz, line 9: Hm, why is the use of the retry command not enough to handle the problems? Also I would not record a "fail" for things we don't need to care about. E.g. temporary issues with repos shouldn't show up as a red box in openQA test failures. How can we keep a bit more simple test code here?

okurz (changes requested):

tests/containers/build.pm
Martchus: I guess having a nice error and also extra hints for known errors is generally quite nice. However, it is a bit annoying that the whole output is now repeated (as it is also already part of the green `wait_serial` before the `record_info`). I'm also not sure whether this is worth the additional code complexity.

Martchus: > I can of course drop the record_info and only die with an appropriate message if that would help?

Or just pass "For details, see previous `wait_serial` step`" instead of `$output` to `record_info`.

okurz: > Would marking them as soft-failure help? In the end it fill still invalidate the rest of the test.

No. Only use "soft-failure" for known issues that are still planned to be worked on. But if it's something that we just accept as given then it should have neither a yellow nor red border.

> The retry is not waiting log enough, it could take more than a few minutes for some cases.

So … configure retry to wait longer?

kalikiana: Maybe for now we should bump the number of attemps (without the other changes). We also discussed some ideas that I'll add to the ticket for a follow-up.

baierjan: I agree. Originally, I increased the sleep time, but increasing the retry count would be more beneficial.

kalikiana (approved):

Martchus (approved):

okurz (approved):

PR #107, opened 2023-02-08
Find job to clone from o3 in a more reliable way

Martchus:
See https://progress.opensuse.org/issues/123864

### DIFF
--- a/tests/osautoinst/start_test.pm
+++ b/tests/osautoinst/start_test.pm
@@ -4,21 +4,17 @@ use testapi;
 use utils;
 
 sub run {
-    # please forgive the hackiness: using the openQA API but parsing the
-    # human-readable output of 'client' to get the most recent job
+    # clone the latest "minimalx" job for the most recent Tumbleweed build with matching architecture
     my $arch       = get_var('ARCH');
     my $ttest      = 'minimalx';
     my $openqa_url = get_var('OPENQA_HOST', 'https://openqa.opensuse.org');
     my $cmd        = <<"EOF";
 set -o pipefail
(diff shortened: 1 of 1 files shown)


Martchus:
Looks like it works: 

grafik

(In this run it cloned https://openqa.opensuse.org/tests/3104794 which is the latest minimalx job from the current Tumbleweed build.)

tests/osautoinst/start_test.pm
```diff
@@ -4,21 +4,17 @@ use testapi;
...
 set -o pipefail
-last_tw_build=\$(OPENQA_CLI_RETRIES=5 openqa-cli api --host $openqa_url assets | jq '.assets | .[] | .name' | sed -n 's/.*Tumbleweed-NET-$arch-Snapshot\\([0-9]\\+\\)-Media.*\$/\\1/p' | sort -n | tail -n 1)
-echo "Last Tumbleweed build on openqa.opensuse.org: \$last_tw_build"
-[ ! -z \$last_tw_build ]
 zypper -n in jq
-resp=\$(OPENQA_CLI_RETRIES=5 openqa-cli api --host $openqa_url jobs version=Tumbleweed scope=relevant arch=$arch build=\$last_tw_build flavor=NET latest=1)
-job_id=\$(echo "\$resp" | jq '.jobs | .[] | select(.test == "$ttest") | .id')
-echo "Job Id: \$job_id"
-if [ -z \$job_id  ]; then echo "Response: \$resp" && false; fi
+resp=\$(OPENQA_CLI_RETRIES=5 openqa-cli api --host $openqa_url jobs jobs version=Tumbleweed scope=relevant arch='$arch' flavor=NET test='$ttest' latest=1)
```
kalikiana, line 14: How is the ` jobs jobs ` supposed to work? Or is it possible it gets ignored silently?

Martchus: Good catch. I suppose it only works because the 2nd jobs is passed as query parameter (with empty value) which is then simply ignored by the server.

okurz (approved):

kalikiana (approved):

PR #108, opened 2023-02-09
install: Allow for 7 retries and a more generous timeout

kalikiana:
See: https://progress.opensuse.org/issues/123496

### DIFF
--- a/tests/install/openqa_worker.pm
+++ b/tests/install/openqa_worker.pm
@@ -5,7 +5,7 @@ use utils;
 
 sub run {
     diag('worker setup');
-    assert_script_run('retry -s 30 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 600);
+    assert_script_run('retry -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 3800);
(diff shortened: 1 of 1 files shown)


Martchus (approved):

okurz (approved):

tests/install/openqa_worker.pm
```diff
@@ -5,7 +5,7 @@ use utils;
 
 sub run {
     diag('worker setup');
-    assert_script_run('retry -s 30 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 600);
+    assert_script_run('retry -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 3800);
```
foursixnine, line 8: Is it me, or are you peeps waiting/hoping that obs packages will be published on time? and making up for it? doesn't the obs provide a message bus to listen to events? Otherwise, you will still face the same sporadic problems over and over.

okurz: openQA-in-openQA tests are triggered by a job on https://jenkins.qa.suse.de/ whenever new packages are published so this should account for that, right? This fix attempt is IIUC not to cover for package inconsistencies or anything but really just for various download problems, same as every openSUSE potentially is also affected by, e.g. https://github.com/openSUSE/zypper/issues/399 which for example I also just hit multiple times trying to upgrade my notebook manually today.

PR #115, opened 2023-04-28
Load python test modules by default with new variable

okurz:
Instead of needing to explicitly specify python in the job templates we
should load python tests by default with a more explanatory variable
name while still offering to skip on problems.

This is inspired by
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/114/files#diff-093d7aec968469f705e116051d4161bd82c22f4add7145791e04e46101881fecR44
showing me that job templates need the explicit variable when it
shouldn't be necessary to reiterate that.

### DIFF
--- a/main.pm
+++ b/main.pm
@@ -77,7 +77,7 @@ elsif (get_var('INSTALL')) {
 # testing from git only tests webui so far
 load_osautoinst_tests() unless check_var('OPENQA_FROM_GIT', 1);
 load_openQA_tests();
-load_python_tests() if get_var('PYTHON');
+load_python_tests() if get_var('LOAD_PYTHON_TEST_MODULES', 1);
 load_shutdown();
 
 1;


Martchus (approved):

kalikiana (approved):
I guess we'll see if this has any side-effects. Making this work by default is definitely nice.

perlpunk:
Yes, we see this breaks openqa_from_containers: https://openqa.opensuse.org/tests/3266519#step/search/2

okurz:
I added `LOAD_PYTHON_TEST_MODULES: '0'` to https://openqa.opensuse.org/admin/job_templates/24 for openqa_from_containers and retriggered the last scheduled product. This triggered jobs 3267052, 3267053, 3267054

https://openqa.opensuse.org/tests/3267054#details shows that the search module is not scheduled anymore for openqa_from_containers so problem fixed.

PR #119, opened 2023-06-05
Use the new LOAD_PYTHON_TEST_MODULES setting

perlpunk:
That was introdued in a3557eca565a9e3487c57614b0ffed1d0a903bd8 but somehow the old PYTHON setting was used in scenario-definitions.yaml

### DIFF
--- a/scenario-definitions.yaml
+++ b/scenario-definitions.yaml
@@ -33,6 +33,7 @@ job_templates:
       UPDATE: "0"
       OPENQA_FROM_GIT: "1"
       OPENQA_CONTAINERS: "1"
+      LOAD_PYTHON_TEST_MODULES: "0"
   openqa_install:
     product: openqa-*-dev-x86_64
     machine: 64bit-2G
@@ -41,4 +42,3 @@ job_templates:
       INSTALL: "1"
       INSTALL_ONLY: "1"
       UPDATE: "0"
-      PYTHON: "1"


perlpunk:
Now this is still failing because the wiorkflow is running on pull_request_target, which only checks out the master branch.
So we need to merge this first, and following PRs will pass.
However, if the workflow does not fetch anything from the PR, it seems kind of useless to me?

baierjan:
> Now this is still failing because the wiorkflow is running on pull_request_target, which only checks out the master branch. So we need to merge this first, and following PRs will pass. However, if the workflow does not fetch anything from the PR, it seems kind of useless to me?

The workflow is spawning new test with `CASEDIR="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY.git#$GITHUB_REF`, so the PR is tested properly (I hope). The `pull_request_target` was selected as a result of https://progress.opensuse.org/issues/128360

baierjan (approved):

perlpunk:
Yes, but it is testing the old `scenario-definitions.yaml` file from master.

baierjan:
> Yes, but it is testing the old `scenario-definitions.yaml` file from master.

Yes. Maybe we can create another workflow for testing the `scenario-definitions.yaml` but that would not be "fork-compatible".

Martchus:
I guess that could be changed by using `SCENARIO_DEFINITIONS_YAML_FILE=https://raw.githubusercontent.com/os-autoinst/os-autoinst-distri-openQA/$GITHUB_REF/.github/workflows/openqa.yml` (instead of `--param-file SCENARIO_DEFINITIONS_YAML=scenario-definitions.yaml`). Then the `- uses: actions/checkout@v3` could likely be dropped as well.

PR #128, opened 2023-06-21
Remove ack installation and usage from test_running

mimi1vx:
ack was here used only as a more advanced grep which isn't necessary. grep has the same functionality and is installed by default.

change in retry -> https://openqa.opensuse.org/tests/3368683 extend time as on slower worker starting scheduled test can take more time.

### DIFF
--- a/tests/osautoinst/test_running.pm
+++ b/tests/osautoinst/test_running.pm
@@ -4,8 +4,7 @@ use testapi;
 use utils;
 
 sub run {
-    assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
-    assert_script_run 'retry -s 30 -r 5 -- openqa-cli api jobs state=running state=done | ack --passthru --color "running|done"', 300;
+    assert_script_run 'retry -s 60 -r 5 -- openqa-cli api jobs state=running state=done | grep -E "running|done"', 600;
     save_screenshot;
     clear_root_console;
 }


Martchus:
I suppose it makes sense to avoid calling zypper just for ack. Is this related to a ticket?

okurz (changes requested):
Please provide proper reasoning in pull requests otherwise we don't know why you do that and what you intend with it. I suggest also to reference the commit that brought in the change originally showing that you understood why something was done in the past and then describe what changed to improve by not doing that anymore

perlpunk:
> ack was here used only as a more advanced grep which isn't necessary. grep has the same functionality and is installed by default.

I think `grep` doesn't provide the same `--passthru` functionality as `ack` - always print complete output even if it doesn't match. See e8439ddf8c4e35947d151726fa83cf8e4ef421df

But #128 fixes the actual problem, so I think this is obsolete.

mimi1vx:
> > ack was here used only as a more advanced grep which isn't necessary. grep has the same functionality and is installed by default.
> 
> I think `grep` doesn't provide the same `--passthru` functionality as `ack` - always print complete output even if it doesn't match. See [e8439dd](https://github.com/os-autoinst/os-autoinst-distri-openQA/commit/e8439ddf8c4e35947d151726fa83cf8e4ef421df)
> 

which isn't important because if isn't match -> tests fail and postfail calls `openqa-cli api jobs` plus `journal` etc ...

okurz:
I'd still prefer to keep ack as is

kalikiana:
> But #128 fixes the actual problem, so I think this is obsolete.

You seem to be referencing this very same PR. What did you mean to refer to?

I can see how avoiding a zypper call makes the test more reliable. So I would be inclined to second the change... but we should simply agree on what we prefer and how hard we want to work around infra and tooling outside our control.

perlpunk:
> > But #128 fixes the actual problem, so I think this is obsolete.
> 
> @perlpunk You seem to be referencing this very same PR. What did you mean to refer to?

sorry, I meant #129 :)

kalikiana:
> > > But #128 fixes the actual problem, so I think this is obsolete.
> > 
> > 
> > @perlpunk You seem to be referencing this very same PR. What did you mean to refer to?
> 
> sorry, I meant #129 :)

Ah, right. This also has the context I was looking for.

@mimi1vx So I would close this one for now. If installing ack is leading to recurring failures I'd say we best have a separate ticket and discuss it as part of the estimation.

PR #130, opened 2023-07-05
Add option for testing with nginx

asdil12:
Ticket: https://progress.opensuse.org/issues/131024

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -16,17 +16,22 @@ sub install_from_repos {
     my $repo = 'openSUSE_' . $repo_suffix{get_required_var('ARCH')};
     $add_repo = "zypper -n ar -p 95 -f obs://devel:openQA/$repo openQA";
     assert_script_run($_) foreach (split /\n/, $add_repo);
-    assert_script_run('retry -s 30 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-local-db"', 600);
-    my $configure = <<'EOF';
-/usr/share/openqa/script/configure-web-proxy
-sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini
-systemctl restart apache2
-systemctl enable --now openqa-webui
-systemctl status --no-pager openqa-webui
-systemctl enable --now openqa-scheduler
-systemctl status --no-pager openqa-scheduler
-EOF
-    assert_script_run($_) foreach (split /\n/, $configure);
+    my $proxy_pkg = (check_var('OPENQA_WEB_PROXY', 'nginx')) ? 'nginx' : '';
+    assert_script_run('retry -s 30 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-local-db '.$proxy_pkg.'"', 600);
+    if (check_var('OPENQA_WEB_PROXY', 'nginx')) {
+        assert_script_run "/usr/share/openqa/script/configure-web-proxy --nginx";
+        assert_script_run "systemctl disable --now apache2";
+        assert_script_run "systemctl restart nginx";
+    }
+    else {
+        assert_script_run "/usr/share/openqa/script/configure-web-proxy";
+        assert_script_run "systemctl restart apache2";
+    }
+    assert_script_run q{sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini};
+    assert_script_run "systemctl enable --now openqa-webui";
+    assert_script_run "systemctl status --no-pager openqa-webui";
+    assert_script_run "systemctl enable --now openqa-scheduler";
+    assert_script_run "systemctl status --no-pager openqa-scheduler";
 }
 
 sub install_from_git {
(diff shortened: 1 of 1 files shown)


tests/install/openqa_webui.pm
```diff
@@ -17,16 +17,21 @@ sub install_from_repos {
...
+    }
+    else {
+        assert_script_run "/usr/share/openqa/script/configure-web-proxy";
+        assert_script_run "systemctl restart apache2";
+    }
+    assert_script_run q{sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini};
+    assert_script_run "systemctl enable --now openqa-webui";
+    assert_script_run "systemctl status --no-pager openqa-webui";
+    assert_script_run "systemctl enable --now openqa-scheduler"
+    assert_script_run "systemctl status --no-pager openqa-scheduler";
```
okurz, line 34: I suggest you extract functions for each block and keep the here-document. I see no reason for that to change

okurz (changes requested):

okurz:
https://github.com/os-autoinst/os-autoinst-distri-openQA/actions/runs/5462802731/jobs/9942654026?pr=130 shows a reference to an incomplete openQA that should be pretty obvious regarding wrong syntax :)

tests/install/openqa_webui.pm
kalikiana: Dominik's code certainly looks cleaner to me.

My only concern would be how that affects code being run in separate calls. Maybe that's actually a good thing since errors will be clearer.

asdil12: The code was already before in several calls using `assert_script_run($_) foreach (split /\n/, $configure);`

perlpunk: Perl 5.26 actualy allows indented here documents: https://perldoc.perl.org/5.26.0/perldelta#Indented-Here-documents

tests/install/openqa_webui.pm
```diff
@@ -17,16 +17,21 @@ sub install_from_repos {
...
+    }
+    else {
+        assert_script_run "/usr/share/openqa/script/configure-web-proxy";
+        assert_script_run "systemctl restart apache2";
+    }
+    assert_script_run q{sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini};
+    assert_script_run "systemctl enable --now openqa-webui";
+    assert_script_run "systemctl status --no-pager openqa-webui";
+    assert_script_run "systemctl enable --now openqa-scheduler";
+    assert_script_run "systemctl status --no-pager openqa-scheduler";
```
perlpunk, line 34: Here is how it cuold look like with an inented here-document:
```suggestion
    assert_script_run($_) foreach (split /\n/, <<~'EOF');
        sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini
        systemctl enable --now openqa-webui
        systemctl status --no-pager openqa-webui
        systemctl enable --now openqa-scheduler
        systemctl status --no-pager openqa-scheduler
        EOF
```

I'm fine with either way.

perlpunk: I do find the here-doc a bit more readable. You see what's going on right away, except for the bit cryptic split above the actual shell code.

tests/install/openqa_webui.pm
```diff
@@ -17,35 +17,41 @@ sub install_from_repos {
...
-sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini
-systemctl restart apache2
-systemctl enable --now openqa-webui
-systemctl status --no-pager openqa-webui
-systemctl enable --now openqa-scheduler
-systemctl status --no-pager openqa-scheduler
-EOF
-    assert_script_run($_) foreach (split /\n/, $configure);
+    if (check_var('OPENQA_WEB_PROXY', 'nginx')) {
+        assert_script_run('retry -s 30 -- sh -c "zypper --no-cd -n in nginx"', 600);
```
okurz, line 21: Instead of this I suggest to use a sub-package that does not install apache but does install nginx. Then we don't have two subsequent zypper install calls but just one.

asdil12: A subpackage that would just recommend nginx and do nothing else would be a bit much overhead.
I will update the PR with a simpler solution.

okurz: Actually we can change to use "openQA-single-instance" in the case of apache2 and have a corresponding "openQA-single-instance-nginx" in the case of nginx. See https://github.com/os-autoinst/openQA/blob/master/dist/rpm/openQA.spec

okurz (approved):

perlpunk (approved):

tests/install/openqa_webui.pm
asdil12: Yes, but here we currently only install the webui. The worker is set up in a separate step.
If we pull in the worker via the single instance rpm, the actual worker test module might have redundant steps then.

This doesn't mean that it doesn't make sense to have a single-instance nginx rpm. I just would not use it here.
Also it would install both nginx and apache unless installed with `--no-recommends` as the openQA rpm recommends apache.

perlpunk:
https://openqa.opensuse.org/tests/3408493#step/dashboard/4 failed with `# Test died: no candidate needle with tag(s) 'desktop-runner, screenlock' matched`, we have a ticket for that

PR #134, opened 2023-07-13
Schedule nginx test in ci

asdil12:


Martchus (approved):
I was just asking for splitting into multiple commits, not a separate PR. However, this is ok as well.

okurz:
I suggest you also manually trigger openQA tests based on that new schedule definition to verify

scenario-definitions.yaml
```diff
@@ -50,3 +50,12 @@ job_templates:
       INSTALL: "1"
       INSTALL_ONLY: "1"
       UPDATE: "0"
+  openqa_install_nginx:
+    product: openqa-*-dev-x86_64
+    machine: 64bit-2G
+    settings:
+      OPENQA_HOST: "http://openqa.opensuse.org"
```
okurz, line 57: why do we need this setting?

okurz (changes requested):

scenario-definitions.yaml
asdil12: idk - I copied the section above which has the same setting.
@baierjan I guess you added that section. Any special reason for this line?

kalikiana (approved):

scenario-definitions.yaml
baierjan: If I recall correctly, the default https protocol failed in some step (maybe due to the weird proxy setup on o3?), this was a workaround recommended by @okurz (https://suse.slack.com/archives/C02AJ1E568M/p1682587547710399?thread_ts=1682587184.320269&cid=C02AJ1E568M)

okurz: Well, that proxy setup is over. Let's try without this setting :) @asdil12 can you please remove it in another commit from all mentions in this file

kalikiana:
@moergifyio rebase

scenario-definitions.yaml
```diff
@@ -43,6 +43,13 @@ job_templates:
...
-      OPENQA_HOST: "http://openqa.opensuse.org"
       MODE: install
       INSTALL_ONLY: "1"
+  openqa_install_nginx:
+    product: openqa-*-dev-x86_64
+    machine: 64bit-2G
+    settings:
+      MODE: install
+      INSTALL_ONLY: "1"
+      UPDATE: "0"
```
okurz, line 54: Remove this line as the `MODE` already covers this

okurz (changes requested):

PR #135, opened 2023-07-20
Make install/update exclusiveness obvious with mode switch

okurz:


main.pm
```diff
@@ -64,10 +64,10 @@ sub load_shutdown() {
 }
 
 # load tests in the right order
-if (get_var('UPDATE')) {
+if (check_var('MODE', 'update') or get_var('UPDATE')) {
```
Martchus, line 67: Looks like "update" is supposed to be the default. So I guess this should be:
```suggestion
if (check_var('MODE', 'update') || check_var('MODE', '')) {
```
I'd also prefer `||` in conditionals (and only use `or` for command-chaining).

main.pm
```diff
@@ -64,10 +64,10 @@ sub load_shutdown() {
...
     load_update_tests();
 }
-elsif (get_var('INSTALL')) {
+elsif (check_var('MODE', 'install') or get_var('INSTALL')) {
```
Martchus, line 70:
```suggestion
elsif (check_var('MODE', 'install') || get_var('INSTALL')) {
```

main.pm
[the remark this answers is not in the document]
okurz: What do you mean? The test variable is set from an outer scope, this is where it's used

okurz: I think you are confusing the use of the second argument of `check_var` here? `check_var('MODE', 'update')` returns true if MODE is set to "update". The backward-alternative evaluation is `get_var('UPDATE')` that evaluates to true if UPDATE is set to a truthy value.

main.pm
okurz: done

perlpunk (approved):

main.pm
Martchus: I was thinking about the case when none of the variables is set and wrote my suggestion so that in this case the update-path is executed. I guess executing none of the special paths is fine as well.

Martchus (approved):

PR #136, opened 2023-07-25
Update ci to retry failing jobs

asdil12:
Ticket: https://progress.opensuse.org/issues/106922


okurz (changes requested):
I doubt that `OPENQA_CLI_RETRIES` will retrigger jobs, it will retry requests, like when a server is not reachable

kalikiana:
As discussed this is not addressing the issue we wanted to deal with

PR #138, opened 2023-07-25
Simplify osautoinst/worker not needing to wait for desktop

okurz:


okurz:
ok, now it's getting ugly. In the worker we called `wait_for_desktop` which is now preventing race conditions with https://github.com/os-autoinst/os-autoinst-distri-openQA/commit/09f85354d36b3d0fa592a3daeba12a6adfd744e4 but after I removed that here now the dashboard module calls instead `ensure_unlocked_desktop` which effectively tries to handle the same but differently.

PR #139, opened 2023-07-28
Add exponential backoff for network-related calls

Martchus:
See https://progress.opensuse.org/issues/133349

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -17,7 +17,7 @@ sub install_from_repos {
     $add_repo = "zypper -n ar -p 95 -f obs://devel:openQA/$repo openQA";
     assert_script_run($_) foreach (split /\n/, $add_repo);
     my $proxy_pkg = (check_var('OPENQA_WEB_PROXY', 'nginx')) ? 'nginx' : '';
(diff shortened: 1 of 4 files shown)


Martchus:
I've tested this locally with some dummy command that would fail. Note that the short form (e.g. `-es 30` instead of `-e -s 30`) does *not* work:

```
$ retry -es 3 -- bash -c 'echo foo && exit 1'
/usr/bin/retry: Zeile 38: 3: Kommando nicht gefunden.
Retrying up to 3 more times after sleeping 3s …
/usr/bin/retry: Zeile 38: 3: Kommando nicht gefunden.
Retrying up to 2 more times after sleeping 0s …
/usr/bin/retry: Zeile 38: 3: Kommando nicht gefunden.
Retrying up to 1 more times after sleeping 0s …
/usr/bin/retry: Zeile 38: 3: Kommando nicht gefunden.
```

Maybe that's something we want to fix within the retry script?

okurz (approved):

perlpunk (approved):

perlpunk:
https://openqa.opensuse.org/tests/3492023#step/test_distribution/3 (problem with Test::Assert) will be fixed soon, see https://bugzilla.opensuse.org/show_bug.cgi?id=1214057
https://openqa.opensuse.org/tests/3492019#step/openqa_webui/13 (problem with Class::Accessor) - I'll have a look

PR #141, opened 2023-07-31
Cleanup+improve complete distribution

okurz:
* Remove obsolete non-install-only modules
* Delete obsolete 'update' tests
* main: Supply extension automatically to loadtest-calls
* Update and consolidate code style

Related progress issue: https://progress.opensuse.org/issues/133025


tests/containers/single_container_webui.pm
```diff
@@ -1,24 +1,23 @@
-use Mojo::Base qw(openQAcoretest);
+use qw(openQAcoretest);
```
perlpunk, line 1: What was the intent here?

okurz: oops, that should still have the Mojo::Base, or at least "base". Fixed

tests/osautoinst/worker.pm
```diff
@@ -1,15 +1,16 @@
...
     wait_for_desktop;
-    send_key 'ctrl-alt-f3';
+    switch_to_root_console;
     assert_screen 'inst-console';
     type_string "root\n";
-    assert_screen 'password-prompt', 10;
-    type_string $testapi::password . "\n";
+    assert_screen 'password-prompt';
+    type_password;
+    send_key 'ret';
```
kalikiana, line 13: I guess what we really want here is this non-existing command:

```suggestion
    enter_password;
```

In that sense I find both the old and your new version equally desirable

kalikiana (approved):

perlpunk (approved):

scenario-definitions.yaml
```diff
@@ -19,34 +19,25 @@ machines:
...
       LOAD_PYTHON_TEST_MODULES: "0"
   openqa_from_bootstrap:
     <<: *common
     settings:
-      <<: *common_settings
       OPENQA_FROM_BOOTSTRAP: "1"
   openqa_install:
     <<: *common
     settings:
-      <<: *common_settings
```
perlpunk, line 47: `settings` is empty here now, so it gets loaded as `null`. I will make a PR to fix it.

PR #143, opened 2023-08-05
Fix scenario-definitions.yaml

perlpunk:
and add a github workflow to validate it in pull requests.

The YAML file got broken in 6616d6b8eae9dc331c491e6bb37a43edfaaa5e90


perlpunk:
The trigger test will fail because it uses scenario-definitions.yaml from master, so we have to ignore that failure here.

okurz (approved):

PR #144, opened 2023-08-09
Change scenario definitions to be usable instead of templates

mimi1vx:
https://progress.opensuse.org/issues/132335

Mergeable after deployment of https://github.com/os-autoinst/openQA/pull/5277

### DIFF
--- a/scenario-definitions.yaml
+++ b/scenario-definitions.yaml
@@ -34,9 +34,11 @@ job_templates:
     <<: *common
     settings:
       OPENQA_FROM_BOOTSTRAP: "1"
-  openqa_install:
-    <<: *common
   openqa_install_nginx:
     <<: *common
     settings:
       OPENQA_WEB_PROXY: 'nginx'
+  openqa_install+publish:
+    settings:
+      PUBLISH_HDD_1: 'opensuse-Tumbleweed-%ARCH%@%MACHINE%-%BUILD%.qcow2'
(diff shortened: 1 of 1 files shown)


scenario-definitions.yaml
```diff
@@ -13,6 +13,8 @@ machines:
       HDDSIZEGB: "20"
       QEMUCPU: host
       QEMURAM: "2048"
+      MODE: install''
```
Martchus, line 16: The duplicated `''` looks wrong.

perlpunk:
The new scenario would now be run for every pull request, right? do we want that?

scenario-definitions.yaml
```diff
@@ -13,6 +13,8 @@ machines:
       HDDSIZEGB: "20"
       QEMUCPU: host
       QEMURAM: "2048"
+      MODE: install
+      INSTALL_ONLY: '1'
```
perlpunk, line 17: These vars aren't used anymore in the code

perlpunk (changes requested):
See comment

mimi1vx:
> The new scenario would now be run for every pull request, right? do we want that?

this is a good question :D I think is little overkill on other side it will use almost all code paths in tests so shows almost all possible problems early

scenario-definitions.yaml
mimi1vx: thx,

kalikiana (approved):
Looks good. Imho we can look into further splitting later if necessary.

perlpunk (approved):

PR #145, opened 2023-09-12
Fix openqa_install+publish scenario

perlpunk:
It needs machine and product.
At least the missing machine will make the next+previous tab not working because `NULL~ is passed to postgres.
Not sure what effect the missing product has.


scenario-definitions.yaml
```diff
@@ -39,6 +39,7 @@ job_templates:
     settings:
       OPENQA_WEB_PROXY: 'nginx'
   openqa_install+publish:
+    <<: *common
```
kalikiana, line 42: How did you notice this was broken? https://openqa.opensuse.org/tests/3570245#next_previous looks all green to me

perlpunk: Ah, I should have added this link: https://openqa.opensuse.org/tests/3570330#next_previous
This only shows one test.

okurz: Do you think we can make the error more prominent in openQA? Overall the test was working quite nicely but the error in openQA is rather subtle, right?

perlpunk: I would first like to find out why the SQL query didn't work with machine set to NULL, because when I tried it manually in psql, the query returned the expected rows.

okurz (approved):

kalikiana (approved):

PR #149, opened 2023-10-23
Adapt post_fail_hook to an incomplete openQA setup

kalikiana:
See: https://progress.opensuse.org/issues/138320

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -14,7 +14,7 @@ sub post_fail_hook {
     if (check_var('OPENQA_FROM_GIT', 1)) {
         send_key 'ctrl-c';     # Stop current command, if any
         assert_script_run 'cd /root/openQA';
-        enter_cmd 'script/openqa-cli api jobs';
+        enter_cmd '[ -e script/openqa-cli ] && script/openqa-cli api jobs';
         save_screenshot;
         enter_cmd 'which sass';
         get_log 'journalctl --pager-end --no-tail --no-pager -u apache2' => 'openqa_services.log';


tests/install/openqa_webui.pm
```diff
@@ -81,4 +81,8 @@ sub run {
     clear_root_console;
 }
 
+sub post_fail_hook {
+    # If the installation failed there is no point in collecting logs of the openQA setup
```
okurz, line 85: But the above code includes starting openQA services when it would make very much sense to collect logs

okurz (changes requested):

perlpunk:
What I find strange is that the jobs triggered by the PR all are missing the install/* tests (and thus failing):
 - https://openqa.opensuse.org/tests/3669735
 - https://openqa.opensuse.org/tests/3669736
 - https://openqa.opensuse.org/tests/3669737
 - https://openqa.opensuse.org/tests/3669738

lib/openQAcoretest.pm
```diff
@@ -14,7 +14,7 @@ sub post_fail_hook {
     if (check_var('OPENQA_FROM_GIT', 1)) {
         send_key 'ctrl-c';     # Stop current command, if any
         assert_script_run 'cd /root/openQA';
-        enter_cmd 'script/openqa-cli api jobs';
+        enter_cmd '[ -e script/openqa-cli ] && script/openqa-cli api jobs';
```
perlpunk, line 17: This will help if the git clone failed, but not if the `cpanm` following that fails.
I guess it's really not that easy.
I'm wondering if there is a general way to make it more obvious that a test is in the post_fail_hook stage.

perlpunk:
> What I find strange is that the jobs triggered by the PR all are missing the install/* tests (and thus failing):

Still not sure but what I see is that https://github.com/kalikiana/os-autoinst-distri-openQA/tree/instl_webui_nofailhook says `This branch is 1 commit ahead, 76 commits behind os-autoinst:master.`.
Maybe rebase?

lib/openQAcoretest.pm
okurz: well, we can just split each test module to be in more specific steps. Then for each module we can define more specific post_fail_hooks

perlpunk:
@kalikiana I would have still been interested in investigating the CI problems by a simple rebase. But I guess we will see in the next PR if it works or not.

perlpunk:
ok, #151 passed, so I assume the reason was that the branch here was based on such an old commit.

PR #150, opened 2023-10-25
CI: Align commit message check to current in openQA

okurz:


okurz:
Now obsolete with inherited check

PR #157, opened 2023-12-20
t: Use consistent 'Mojo::Base' instead of strict+warnings

okurz:
Same as suggested in our documentation we should use Mojo::Base
consistently in all openQA test modules.

Motivated by:
* https://github.com/os-autoinst/os-autoinst-distri-example/pull/28

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -2,7 +2,6 @@ package utils;
 
 use Mojo::Base 'Exporter', -signatures;
 use Exporter;
-use strict;
 use testapi;
 use File::Basename qw(basename);
 
--- a/tests/containers/build.pm
+++ b/tests/containers/build.pm
@@ -1,5 +1,4 @@
-use strict;
-use base 'openQAcoretest';
+use Mojo::Base 'openQAcoretest';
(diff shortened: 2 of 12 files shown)


kalikiana (approved):

Martchus (approved):

okurz:
ok, I think the problem is that in https://github.com/os-autoinst/os-autoinst-distri-openQA/actions/runs/7277050058/job/19828218115?pr=157#step:5:11 `(tr '/' ':' <<<"$GH_REPO"` is done but not on GH_REF which has a / for me

perlpunk:
Yeah, I thought the same, see #158

okurz:
@mergifyio rebase

perlpunk:
Now https://openqa.opensuse.org/tests/3831187#step/boot/4 fails, with the same error we see in our regular tests: https://openqa.opensuse.org/tests/3831958#step/boot/4 but which we aren't notified of because no emails :-/

perlpunk:
https://progress.opensuse.org/issues/152889
I added a needle, and retriggered the test

PR #159, opened 2024-01-05
Allow nested multi machine tests

asdil12:
https://progress.opensuse.org/issues/138302

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -1,16 +1,13 @@
 package openQAcoretest;
 use Mojo::Base 'basetest', -signatures;
 use testapi;
-use utils qw(switch_to_root_console);
+use utils qw(switch_to_root_console get_log);
 
 
-sub get_log ($cmd, $name) {
-    my $ret = script_run "$cmd | tee $name";
-    upload_logs($name) unless $ret;
-}
-
-sub post_fail_hook {
+sub post_fail_hook ($self) {
     switch_to_root_console;
+    # we can't upload logs if the multimachine OVS bridge in the SUT has the same IP as the openQA-worker host
+    assert_script_run 'ip a del 10.0.2.2/15 dev br1';
     if (get_var('OPENQA_FROM_GIT')) {
         send_key 'ctrl-c';     # Stop current command, if any
         assert_script_run 'cd /root/openQA';
@@ -24,7 +21,7 @@ sub post_fail_hook {
         enter_cmd 'openqa-cli api jobs';
         save_screenshot;
         get_log 'journalctl --pager-end --no-tail --no-pager -u apache2 -u nginx -u openqa-scheduler -u openqa-websockets -u openqa-webui -u openqa-worker@1' => 'openqa_services.log';
-        get_log 'cat /var/lib/openqa/pool/1/autoinst-log.txt /var/lib/openqa/testresults/*/*/autoinst-log.txt' => 'autoinst-log.txt';
+        get_log '(cat /var/lib/openqa/pool/1/autoinst-log.txt /var/lib/openqa/testresults/*/*/autoinst-log.txt ||:)' => 'autoinst-log.txt';
     }
 }
 
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -5,7 +5,12 @@ use Exporter;
 use testapi;
 use File::Basename qw(basename);
 
-our @EXPORT = qw(clear_root_console switch_to_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log prepare_firefox_autoconfig disable_packagekit);
+our @EXPORT = qw(get_log clear_root_console switch_to_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log prepare_firefox_autoconfig disable_packagekit);
+
+sub get_log ($cmd, $name) {
+    my $ret = script_run "$cmd | tee $name";
+    upload_logs($name) unless $ret;
+}
 
 sub clear_root_console {
     enter_cmd 'clear';
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -35,23 +35,16 @@ sub handle_gui_password {
 sub wait_for_desktop {
     assert_screen([qw/boot-menu openqa-desktop/]);
     send_key 'ret' if match_has_tag('boot-menu');
-    for (1..3) {
-        assert_screen 'openqa-desktop', 500;
-        if (match_has_tag('openqa-desktop-locked')) {
-            send_key 'esc';
-            handle_gui_password;
-        }
-        elsif (match_has_tag('openqa-desktop-login')) {
-            assert_and_click 'openqa-desktop-login';
-            handle_gui_password;
-        }
-        elsif (match_has_tag('openqa-desktop-gnome-auth-required')) {
-            assert_and_click 'openqa-desktop-gnome-auth-required';
-        }
-        else {
-            last;
-        }
+    assert_screen 'openqa-desktop', 500;
+    return if match_has_tag('generic-desktop');
+    if (match_has_tag('openqa-desktop-locked')) {
+        send_key 'esc';
     }
+    elsif (match_has_tag('openqa-desktop-login')) {
+        assert_and_click 'openqa-desktop-login';
+    }
+    handle_gui_password;
+    assert_screen 'generic-desktop';
 }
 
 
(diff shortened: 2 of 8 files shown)


lib/openQAcoretest.pm
```diff
@@ -1,16 +1,13 @@
...
+use utils qw(switch_to_root_console get_log);
 
 
-sub get_log ($cmd, $name) {
-    my $ret = script_run "$cmd | tee $name";
-    upload_logs($name) unless $ret;
-}
-
-sub post_fail_hook {
+sub post_fail_hook($self) {
```
Martchus, line 7:
```suggestion
sub post_fail_hook ($self) {
```

tests/install/openqa_worker.pm
```diff
@@ -15,13 +15,20 @@ key = 1234567890ABCDEF
 secret = 1234567890ABCDEF
 EOF
 ');
+    if (get_var('FULL_MM_TEST')) {
+		# add tap class to worker config
```
Martchus, line 19:
```suggestion
        # add tap class to worker config
```

lib/openQAcoretest.pm
```diff
@@ -1,16 +1,13 @@
...
     switch_to_root_console;
+    # we can't upload logs if the multimachine OVS bridge in the SUT has the same IP as the openQA-worker host
+    script_run 'ip a del 10.0.2.2/15 dev br1';
```
okurz, line 10: Can you please use `assert_script_run`? This shouldn't fail

lib/utils.pm
```diff
@@ -40,10 +45,12 @@ sub wait_for_desktop {
         if (match_has_tag('openqa-desktop-locked')) {
             send_key 'esc';
             handle_gui_password;
+            last if (check_screen('generic-desktop', 10));
```
okurz, line 48: Please use a multi-tag `assert_screen` with `match_has_tag` instead of `check_screen` with non-zero timeout to prevent introducing any timing dependant behaviour, to save test execution time as well as state more explicitly from the testers point of view what are the expected alternatives. For example:

```perl
assert_screen([qw(yast2_console-finished yast2_missing_package)]);
if (match_has_tag('yast2_missing_package')) {
    send_key 'alt-o';  # confirm package installation
    assert_screen 'yast2_console-finished';
}
```

See https://github.com/os-autoinst/os-autoinst-distri-opensuse/blob/master/CONTRIBUTING.md#coding-style for more details

tests/osautoinst/test_running.pm
```diff
@@ -2,19 +2,37 @@ use Mojo::Base 'openQAcoretest', -signatures;
...
+
+sub post_fail_hook($self) {
     $self->SUPER::post_fail_hook;
     script_run 'lsmod | grep kvm';
     save_screenshot;
     get_log('grep --color -z -E "(vmx|svm)" /proc/cpuinfo' => 'cpuinfo');
     assert_script_run 'grep --color -z -E "(vmx|svm)" /proc/cpuinfo', fail_message => 'Machine does not support nested virtualization, please enable in worker host';
+    if (get_var('FULL_MM_TEST')) {
+        $self->upload_mm_logs();
+    }
```
okurz, line 35:
```suggestion
    $self->upload_mm_logs() if get_var('FULL_MM_TEST');
```

tests/install/openqa_worker.pm
```diff
@@ -15,13 +15,20 @@ key = 1234567890ABCDEF
...
     my $worker_setup = <<'EOF';
-systemctl start openqa-worker@1
+systemctl enable --now os-autoinst-openvswitch
+systemctl status --no-pager os-autoinst-openvswitch
+systemctl enable --now openqa-worker@1
 systemctl status --no-pager openqa-worker@1
-systemctl enable openqa-worker@1
 EOF
     assert_script_run($_) foreach (split /\n/, $worker_setup);
+    assert_script_run "systemctl enable --now openqa-worker@2" if (get_var('FULL_MM_TEST'));
```
okurz, line 31:
```suggestion
    assert_script_run "systemctl enable --now openqa-worker@2" if get_var('FULL_MM_TEST');
```

tests/install/openqa_worker.pm
```diff
@@ -15,13 +15,20 @@ key = 1234567890ABCDEF
 secret = 1234567890ABCDEF
 EOF
 ');
+    if (get_var('FULL_MM_TEST')) {
+		# add tap class to worker config
+        my $arch = get_var('ARCH');
```
okurz, line 20:
```suggestion
        my $arch = get_required_var('ARCH');
```

tests/install/openqa_worker.pm
```diff
@@ -5,7 +5,7 @@ use utils;
 sub run {
     diag('worker setup');
     disable_packagekit;
-    assert_script_run('retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker"', 3800);
+    assert_script_run('retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in openQA-worker os-autoinst-openvswitch"', 3800);
```
okurz, line 8: Why is it necessary to pull in another package? We should make sure that our packages as well as documentation bring along everything that is necessary to run multi-machine tests. Actually os-autoinst-setup-multi-machine according to our docs pull that in so that should be used instead

lib/utils.pm
```diff
@@ -40,10 +45,12 @@ sub wait_for_desktop {
...
         }
         elsif (match_has_tag('openqa-desktop-login')) {
             assert_and_click 'openqa-desktop-login';
             handle_gui_password;
+            last if (check_screen('generic-desktop', 10));
```
okurz, line 53: same as above

okurz (changes requested):

lib/utils.pm
asdil12: I did this exactly like this on purpose.
This is a hard requirement to fix race conditions when logging in, that would appear nearly always when I run this test locally.

The timeout is intended and wanted, and I only use a single needle because this is the needle that is expected after performing login and it won't introduce a big delay in the case of success.

The problem is that after login, for some second or so we still get the old screen content.

lib/utils.pm
asdil12: same reason as above.

lib/utils.pm
okurz: not resolved

okurz (changes requested):

lib/utils.pm
okurz: The "wait_for_desktop" story has more context. There are already multiple tickets related to that and the near-duplicate code that we have in a similar manner in os-autoinst-distri-opensuse. Introducing more differences is not a good idea.

okurz: @baierjan didn't you recently worked on related code and have some suggestion here?

asdil12:
VR for changes in `wait_for_desktop`:
image

Martchus (approved):

okurz (commented):
LGTM. Still draft so not approving

lib/utils.pm
baierjan: Yes, I did. If I remember correctly, I did not need to touch the code and my problem was solved with https://github.com/os-autoinst/os-autoinst-needles-openQA/pull/23.

asdil12:
With https://github.com/os-autoinst/openQA/pull/5426 merged, the results look much better:
https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=dheidler_mm3_ping&distri=openqa&groupid=38

okurz:
> With [os-autoinst/openQA#5426](https://github.com/os-autoinst/openQA/pull/5426) merged, the results look much better: https://openqa.opensuse.org/tests/overview?version=Tumbleweed&build=dheidler_mm3_ping&distri=openqa&groupid=38

Nice. There are two failures though:
* https://openqa.opensuse.org/tests/3874423#step/test_running/6
* https://openqa.opensuse.org/tests/3874464#step/test_running/6

asdil12:
Both are agetty showing unexpectedly on the serial tty interfering with the test.
I would consider that as a general issue that could appear in any openqa test - so unrelated to this testsuite.

See https://openqa.opensuse.org/tests/3874464#comments

> Nice. There are two failures though:
> 
> * https://openqa.opensuse.org/tests/3874423#step/test_running/6
> * https://openqa.opensuse.org/tests/3874464#step/test_running/6

asdil12:
Handling the agetty issues in https://progress.opensuse.org/issues/153766

okurz (approved):

PR #160, opened 2024-01-17
Fix selection of 4G RAM for MM test

asdil12:
https://progress.opensuse.org/issues/138302

### DIFF
--- a/scenario-definitions.yaml
+++ b/scenario-definitions.yaml
@@ -14,10 +14,21 @@ machines:
       QEMUCPU: host
       QEMURAM: "2048"
       WORKER_CLASS: qemu_x86_64
+  64bit-4G:
+    backend: qemu
+    settings:
+      HDDSIZEGB: "20"
+      QEMUCPU: host
+      QEMURAM: "4096"
+      WORKER_CLASS: qemu_x86_64
 
 .common: &common
   product: openqa-*-dev-x86_64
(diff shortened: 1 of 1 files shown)


scenario-definitions.yaml
```diff
@@ -43,10 +54,9 @@ job_templates:
       DESKTOP: minimalx
       OPENQA_WEB_PROXY: 'nginx'
   openqa_install_multimachine:
-    <<: *common
+    <<: *common4G
     settings:
       DESKTOP: minimalx
-      QEMURAM: "4096"
```
Martchus, line 49: So specifying the RAM on job template level didn't work?

asdil12: no - for whatever reason it doesn't win against the machine. Maybe this is something specific to scenario definitions.
https://openqa.opensuse.org/tests/3875368#settings

Martchus (approved):

Martchus:
Considering the failing CI checks it looks like the config is not effective yet: https://openqa.opensuse.org/tests/3875362#settings

scenario-definitions.yaml
kalikiana: So + syntax wouldn't work? Maybe this should just be documented, or put in a ticket if it might be a bug.

okurz (approved):

PR #163, opened 2024-02-16
Bump timeout for `os-autoinst-setup-multi-machine`

Martchus:
The default timeout of 90 seconds seems a little bit too short, see https://progress.opensuse.org/issues/155173#note-11. Adding 30 seconds should give us enough headroom to avoid running into timeouts. Note that in production we even have a timeout of 20 minutes just for starting `os-autoinst-openvswitch.service` but hopefully 2 minutes will be enough for our test setup.

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -39,7 +39,7 @@ sub install_from_repos {
 
 sub install_from_git {
     assert_script_run($_, 600) foreach (split /\n/, <<~'EOF');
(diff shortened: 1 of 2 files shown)


Martchus:
The CI fails because of the module `worker` but that's currently a general problem. The other failure in `openqa_webui` (https://openqa.opensuse.org/tests/3943318#step/openqa_webui/35) also looks unrelated. I guess I can look into those failures as well.

perlpunk (approved):

okurz:
I would bump the timeout even way further. The question is: imagine the setup fails sometimes even with the timeout of 120s. What would you do? It's not unrealistic to assume that we simply accept a longer timeout, maybe even 5m as "still acceptable"

PR #164, opened 2024-02-16
Prevent running into DPMS timeout

Martchus:
Based on #163

### DIFF
--- a/tests/install/boot.pm
+++ b/tests/install/boot.pm
@@ -1,8 +1,20 @@
 use Mojo::Base 'openQAcoretest';
+use testapi;
 use utils;
+
+sub _prevent_inactivity_timeout {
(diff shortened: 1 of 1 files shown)


tests/install/boot.pm
```diff
@@ -1,8 +1,14 @@
...
+use testapi;
 use utils;
 
+use OpenQA::Wheel::Launcher 'start_gui_program';
+
+sub _prevent_inactivity_timeout { start_gui_program('xset s off -dpms', valid => 1) }
+
 sub run {
     wait_for_desktop;
+    _prevent_inactivity_timeout;
```
okurz, line 11: I would prefer that the boot module really only does what it says and stays that simple. How about a separate test module where you execute that?

Martchus:
When testing this locally via the developer mode I noticed that this command does not help to disable XScreenSaver in this environment. Maybe the approach taken in #166 will help.

PR #165, opened 2024-02-19
Allow the post_fail_hook ip command to fail

perlpunk:
It may fail in case this IP is not actually set on the bridge.
Also the ctrl-c needs to happen before, because if the

    grep ... tail -f -n0 nohup.out
    
fails, it's not killed. Since ctrl-c shouldn't do any harm, I guess we can
do it regardless of which test we are running.

This ensures that the post_fail_hook can be completed.

Also this fixes sporadic failures in openqa_from_git: https://openqa.opensuse.org/tests/3946911#step/openqa_webui/36

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -6,10 +6,10 @@ use utils qw(switch_to_root_console get_log);
 
 sub post_fail_hook ($self) {
     switch_to_root_console;
+    send_key 'ctrl-c';     # Stop current command, if any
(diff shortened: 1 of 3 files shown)


okurz (approved):

asdil12 (approved):

Martchus (approved):

perlpunk:
Added another commit that improves the grepping of nohup.out

okurz (approved):

PR #168, opened 2024-02-27
Specify groupid when cloning jobs to prevent cloning VRs

asdil12:
This will prevent cloning jobs that were manually scheduled
by devs outside of the jobgroup.

Fixes issues like https://openqa.opensuse.org/tests/3966516

### DIFF
--- a/tests/osautoinst/start_test.pm
+++ b/tests/osautoinst/start_test.pm
@@ -2,13 +2,13 @@ use Mojo::Base 'openQAcoretest', -signatures;
 use testapi;
 use utils;
 
-sub fetch_job_id($ttest, $flavor, $openqa_url) {
+sub fetch_job_id($groupid, $ttest, $flavor, $openqa_url) {
     # Stores the job id of the latest $ttest job for the most recent Tumbleweed build with matching architecture in $job_id on the shell on the SUT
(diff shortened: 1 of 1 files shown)


tests/osautoinst/start_test.pm
```diff
@@ -2,13 +2,13 @@ use Mojo::Base 'openQAcoretest', -signatures;
 use testapi;
 use utils;
 
-sub fetch_job_id($ttest, $flavor, $openqa_url) {
+sub fetch_job_id($groupid, $ttest, $flavor, $openqa_url) {
     # Stores the job id of the latest $ttest job for the most recent Tumbleweed build with matching architecture in $job_id on the shell on the SUT
```
okurz, line 6:
```suggestion
sub fetch_job_id($ttest, $flavor, $openqa_url, $groupid = 1) {
    # Stores the job id of the latest $ttest job for the most recent Tumbleweed build with matching architecture in $job_id on the shell on the SUT
```

and leave out the "1" when calling

perlpunk (approved):

Martchus (approved):

tests/osautoinst/start_test.pm
okurz: ok

PR #169, opened 2024-03-12
Set explicit timeout on fetch_job_id

b10n1k:
Increase timeout to prevent timeout on openqa-cli requests.

https://progress.opensuse.org/issues/156919


Martchus (approved):

Martchus (commented):
Maybe this timeout could even be higher and combined with introducing exponential backoff in `openqa-cli` itself (as suggested in the ticket).

okurz (changes requested):
As explained on http://open.qa/api/testapi/#_assert_script_run the default timeout is already 90s so this change is not doing what you state it would do to "increase the timeout"

PR #170, opened 2024-04-09
Consolidate package installs with retries

kalikiana:
See: https://progress.opensuse.org/issues/158553

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -5,11 +5,16 @@ use Exporter;
 use testapi;
 use File::Basename qw(basename);
 
-our @EXPORT = qw(get_log clear_root_console switch_to_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log prepare_firefox_autoconfig disable_packagekit);
+our @EXPORT = qw(get_log install_packages clear_root_console switch_to_root_console switch_to_x11 wait_for_desktop ensure_unlocked_desktop wait_for_container_log prepare_firefox_autoconfig disable_packagekit);
 
 sub get_log ($cmd, $name) {
     my $ret = script_run "$cmd | tee $name";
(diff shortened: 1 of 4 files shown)


lib/utils.pm
```diff
@@ -12,6 +12,11 @@ sub get_log ($cmd, $name) {
     upload_logs($name) unless $ret;
 }
 
+sub install_packages($packages, $timeout = undef) {
+    $timeout //= 600;
+    assert_script_run(qq{retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && zypper --no-cd -n in $packages"}, $timeout);
```
kalikiana, line 17: Let's see how well this works. I wasn't sure about `-n` and the timeout(s). Since this is racy by design I spawned 100 anyway: https://openqa.opensuse.org/tests/overview?build=poo%23158553

kalikiana:
image

    Cannot find device "br1"

I don't understand this error. It seems to happen after removing `xscreensaver`. This doesn't seem like it would be caused by my change.

perlpunk:
https://openqa.opensuse.org/tests/4073263#step/openqa_webui/2
```
# Test died: Undefined subroutine &openqa_webui::install_packages called at /var/lib/openqa/pool/18/os-autoinst-distri-openQA//tests/install/openqa_webui.pm line 63.
```

lib/utils.pm
okurz: Can you please open an issue on github.com/openSUSE/zypper/ and reference that here for our "multi-layer workaround"?

okurz (approved):

kalikiana:
Verification runs look better now: https://openqa.opensuse.org/tests/overview?distri=openqa&build=poo%23158553_n2&version=Tumbleweed

Martchus (approved):

PR #175, opened 2024-05-28
Deduplicate machine settings in scenario def

asdil12:
Use YAML merge keys.
Ticket: https://progress.opensuse.org/issues/160209


kalikiana (approved):

Martchus:
You didn't mention how you tested this? (I guess I would just call `yq` on the old and new version and if there's no difference than everything is good.)

perlpunk (approved):

PR #178, opened 2024-06-13
codespace: Run chown as postStartCommand

perlpunk:
Issue: https://progress.opensuse.org/issues/87695

It seems when run as postCreateCommand it is only run on creation, but not on recreation (when restarting a codespace). So currently after a restart the permissions are set back to what they were before.


perlpunk:
ci failures because of https://progress.opensuse.org/issues/162242 and the `sed` issue - don't know which ticket.
https://openqa.opensuse.org/tests/4270742
https://openqa.opensuse.org/tests/4270745
https://openqa.opensuse.org/tests/4270747

Shouldn't prevent us from merging as this is unrelated from this PR

kalikiana (approved):

PR #180, opened 2024-06-17
Run sass with --trace

perlpunk:
Issue: https://progress.opensuse.org/issues/162311

In case there is an error, e.g.

    TypeError: String can't be coerced into Integer
      Use --trace for backtrace.

we don't get a line number or the input. Nothing.

--trace will hopefully help us find the source of this sporadic error.

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -55,10 +55,13 @@ sub install_from_git {
     systemctl restart apache2 || systemctl status --no-pager apache2
     mkdir -p /var/lib/openqa/db
     EOF
+    my $sass = script_output 'perldoc -l Mojolicious::Plugin::AssetPack::Pipe::Sass';
+    script_run q{perl -pi -wE 's/\Q = (qw(sass -s)/ = (qw(sass -s --trace)/' } . $sass;
+    script_run qq{grep "sass -s" $sass};
     script_run('env OPENQA_CONFIG=etc/openqa nohup script/openqa daemon &', 0);
     diag('Wait until the server is responsive');
     assert_script_run('grep -qP "Listening at.*(127.0.0.1|localhost)" <(tail -F nohup.out) ', 600);
-    get_log 'cat nohup.out' => 'openqa_nohup_out.txt';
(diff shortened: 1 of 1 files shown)


Martchus (approved):
Looks like you found the right place/file to patch :-)

perlpunk:
https://openqa.opensuse.org/tests/4275646#step/openqa_webui/5 still failing because of the `sed` issue, otherwise passes

tests/install/openqa_webui.pm
```diff
@@ -55,10 +55,13 @@ sub install_from_git {
     systemctl restart apache2 || systemctl status --no-pager apache2
     mkdir -p /var/lib/openqa/db
     EOF
+    my $sass = script_output 'perldoc -l Mojolicious::Plugin::AssetPack::Pipe::Sass';
+    script_run q{perl -pi -wE 's/\Q = (qw(sass -s)/ = (qw(sass -s --trace)/' } . $sass;
+    script_run qq{grep "sass -s" $sass};
```
okurz, line 60: It's bad practice to use `script_run` without checking the exit code. How about

```suggestion
    assert_script_run q{perl -pi -wE 's/\Q = (qw(sass -s)/ = (qw(sass -s --trace)/' } . $sass;
    assert_script_run qq{grep "sass -s" $sass};
```

but in general I would prefer if our code can be as simple as possible resembling what users would need to do. This is quite low-level. How about a Makefile target "prepare-assets" or something that is then called here?

tests/install/openqa_webui.pm
```diff
@@ -55,10 +55,13 @@ sub install_from_git {
...
     script_run('env OPENQA_CONFIG=etc/openqa nohup script/openqa daemon &', 0);
     diag('Wait until the server is responsive');
     assert_script_run('grep -qP "Listening at.*(127.0.0.1|localhost)" <(tail -F nohup.out) ', 600);
-    get_log 'cat nohup.out' => 'openqa_nohup_out.txt';
+    upload_logs('nohup.out', log_name => 'openqa_nohup_out.txt');
```
okurz, line 64: that looks like it would be better in a separate commit

okurz (changes requested):

PR #181, opened 2024-06-17
Upload log with .txt suffix

perlpunk:
This way openQA will display it directly instead of downloading it.

Related issue: https://progress.opensuse.org/issues/162311


perlpunk:
https://openqa.opensuse.org/tests/4275656#step/openqa_webui/5 fails because of the `sed` issue, otherwise passes.
This is getting annoying

okurz (approved):

PR #183, opened 2024-06-19
Add optional install from repos for all branches

okurz:
By default this will still add devel:openQA only for the test branch
that installs from packages but by setting the test variable
ADD_OPENQA_REPO to a thruthy value one can still add the repo in other
cases, e.g. to install openQA-bootstrap also from devel:openQA.
Additionally the addition of the repository can be disabled by setting
ADD_OPENQA_REPO to false.

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -1,11 +1,9 @@
 use Mojo::Base 'openQAcoretest';
 use testapi;
-use utils qw(install_packages get_log clear_root_console);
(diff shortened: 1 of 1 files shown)


okurz:
openqa: https://openqa.opensuse.org/tests/4282524 ADD_OPENQA_REPO=1 TEST=+-add_openqa_repo

okurz:
-> https://openqa.opensuse.org/tests/4282910

Martchus (approved):

PR #186, opened 2024-07-12
Add debug output in case 'zypper ref' fails

okurz:
Called

```
openqa-clone-job --repeat 2000 --within-instance https://openqa.opensuse.org/tests/4336200 PUBLISH_HDD_1= INCLUDE_MODULES=boot,prepare,openqa_webui
```

-> https://openqa.opensuse.org/tests/overview?flavor=dev&build=okurz%3Aos-autoinst-distri-openQA%23feature%3Adebug_zypper_repo_refresh

While tests started I triggered a rebuild on https://build.opensuse.org/package/live_build_log/devel:openQA/openQA/openSUSE_Tumbleweed/x86_64

Related progress issue: https://progress.opensuse.org/issues/163112

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -1,6 +1,6 @@
-use Mojo::Base 'openQAcoretest';
+use Mojo::Base 'openQAcoretest', -signatures;
 use testapi;
-use utils qw(install_packages clear_root_console);
+use utils qw(install_packages clear_root_console get_log);
 
 
 sub add_repo {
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

okurz:
Now unfortunately all 2k jobs in https://openqa.opensuse.org/tests/overview?flavor=dev&build=okurz%3Aos-autoinst-distri-openQA%23feature%3Adebug_zypper_repo_refresh passed , that was unexpected :D

tests/install/openqa_webui.pm
```diff
@@ -102,4 +102,13 @@ sub run {
     clear_root_console;
 }
 
+sub post_fail_hook ($self) {
+    get_log 'tail -n 400 /var/log/zypper.log' => 'zypper.log.txt';
+    get_log 'ls -la /var/cache/zypp/raw/openQA/repodata/' => 'repodata.log.txt';
+    # "-vvv" added due to https://progress.opensuse.org/issues/163112
+    assert_script_run('zypper -n --gpg-auto-import-keys -vvv ref', timeout => 600);
```
DimStar77, line 109: this has the potential to abort the post_fail_hook (and thus not reaching `$self->SUPER::post_fail_hook`)

okurz: Yes. But possibly the parent post fail hook does not tell us more then. I guess we should be ok to see what happens about that. We monitor the execution of all openQA-in-openQA tests in the tools team and I expect that we see the effect.

DimStar77: we don't get all the logs uploaded without the parent fail_hook - sometimes they can have nice information

PR #189, opened 2024-07-29
Improve uploading of openqa logs

perlpunk:
Issue: https://progress.opensuse.org/issues/163787


okurz (approved):
I guess I would prefer a solution in pure shell, not that much mixing shell and perl but I can't think of a good specific code suggestion right now so approved as-is

Martchus (approved):

PR #190, opened 2024-07-30
Also collect logs from GRU service on error

okurz:
Related progress issue: https://progress.opensuse.org/issues/163784


perlpunk (commented):
Could have been a simple suggestion to #189?

perlpunk:
I included it now in #189

Martchus:
Then let's merge #189 instead.

PR #192, opened 2024-08-02
Debug errors related to Git cloning

Martchus:
See particular commit messages and https://progress.opensuse.org/issues/163784

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -31,6 +31,7 @@ sub upload_openqa_logs {
     chomp(my @logs = qx{find /var/lib/openqa/pool/1/autoinst-log.txt /var/lib/openqa/testresults/*/*/autoinst-log.txt});
     @logs and get_log "(cat @logs ||:)" => 'autoinst-log.txt';
     get_log '(find /var/lib/openqa/pool/ /var/lib/openqa/testresults/ ||:)' => 'find.txt';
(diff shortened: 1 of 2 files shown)


perlpunk (approved):

Martchus:
Looks like this doesn't work: https://openqa.opensuse.org/tests/4376629#step/test_running/21

lib/openQAcoretest.pm
```diff
@@ -31,6 +31,7 @@ sub upload_openqa_logs {
     chomp(my @logs = qx{find /var/lib/openqa/pool/1/autoinst-log.txt /var/lib/openqa/testresults/*/*/autoinst-log.txt});
     @logs and get_log "(cat @logs ||:)" => 'autoinst-log.txt';
     get_log '(find /var/lib/openqa/pool/ /var/lib/openqa/testresults/ ||:)' => 'find.txt';
+    get_log q|sudo -u geekotest /usr/share/openqa/script/openqa eval -V 'my $jobs = app->minion->jobs; my @r; while (my $j = $jobs->next) { push @r, $j->{result} }; \@r'| => 'openqa_minion_results.txt';
```
kalikiana, line 34: The file should show up in the assets now, right? Seems like it doesn't? Looking at https://openqa.opensuse.org/tests/4376707#downloads for instance

perlpunk: Only
1. in case of a failure when the post_fail_hook is run
2. in multimachine tests where we always upload the logs

Actually I think it could be useful to upload the logs in every case and not just failures.
Can be useful when comparing passed and failed jobs

Martchus: Exactly, and thanks to case 2 we can verify that it actually works: https://openqa.opensuse.org/tests/4376711/logfile?filename=test_running-openqa_minion_results.txt

kalikiana (approved):

PR #195, opened 2024-08-06
Add even more diagnostics for zypper ref

andrii-suse:
1. Single `zypper ref` is expected to fail when metadata is concurrenlty changing on the server. Thus adding back `retry` to tackle that.
2. The goal is to gave log from zypper at least for two subsequent failures, to confirm how it fetches the metadata. Thus trying to collect more logs.
3. For some reasons repodata.log.txt was empty. Thus try to collect more info for it as well

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -13,7 +13,7 @@ sub add_repo {
     my $repo = 'openSUSE_' . $repo_suffix{get_required_var('ARCH')};
     my $repo_url = get_var('OPENQA_REPO_URL', "obs://devel:openQA/$repo");
     assert_script_run("zypper -n ar -p 95 -f '$repo_url' openQA");
-    assert_script_run('zypper -n --gpg-auto-import-keys ref', timeout => 600);
+    assert_script_run('retry -e -s 30 -r 7 -- zypper -n --gpg-auto-import-keys ref', timeout => 4000);
 }
 
 sub install_from_pkgs {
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

tests/install/openqa_webui.pm
```diff
@@ -103,10 +103,11 @@ sub run {
...
 sub post_fail_hook ($self) {
-    get_log 'tail -n 400 /var/log/zypper.log' => 'zypper.log.txt';
-    get_log 'ls -la /var/cache/zypp/raw/openQA/repodata/' => 'repodata.log.txt';
-    # "-vvv" added due to https://progress.opensuse.org/issues/163112
-    assert_script_run('zypper -n --gpg-auto-import-keys -vvv ref', timeout => 600);
+    get_log 'tail -n 50000 /var/log/zypper.log' => 'zypper.log.txt';
+    get_log 'ls -la /var/cache/zypp/raw/' => 'repodata.log.txt';
+    assert_script_run('zypper -n --gpg-auto-import-keys ref', timeout => 600);
+    get_log 'tail -n 800 /var/log/zypper.log' => 'zypper1.log.txt';
+    get_log 'ls -lRa /var/cache/zypp/raw/' => 'repodata1.log.txt'
```
perlpunk, line 110:
```suggestion
    get_log 'ls -lRa /var/cache/zypp/raw/' => 'repodata1.log.txt';
```

andrii-suse:
so it looks the root cause is that download.o.o redirects request repomd* files, which should not happen. I will analyze  that redirect, no additional info is needed atm.

andrii-suse:
The handling of repomd.xml should work properly now, but `zypper refresh` still might fail without retry if files are being changed concurrently on the server. So I think it worth to merge this PR to have more data in case of further failures.

perlpunk (approved):

okurz (approved):

PR #196, opened 2024-08-09
Retry openqa-bootstrap if it fails within the test

kalikiana:
The script can fail due to temporary failures or
race conditions whilst installing packages.

See: https://progress.opensuse.org/issues/165024

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -79,7 +79,8 @@ sub install_containers {
 
 sub install_from_bootstrap {
     install_packages('openQA-bootstrap');
-    assert_script_run('skip_suse_specifics=1 skip_suse_tests=1 /usr/share/openqa/script/openqa-bootstrap', timeout => 1200);
+    # The maximum of the retry is 3810 seconds
+    assert_script_run(qq{retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && skip_suse_specifics=1 skip_suse_tests=1 /usr/share/openqa/script/openqa-bootstrap"}, 4000);
 }
 
 sub run {


okurz (changes requested):
Can you comment why not what the ticket https://progress.opensuse.org/issues/165024 suggested to do as first alternative which is to do the retry within openqa-bootstrap?

kalikiana:
> Can you comment why not what the ticket https://progress.opensuse.org/issues/165024 suggested to do as first alternative which is to do the retry within openqa-bootstrap?

I don't understand the question.

#5840 is about the second suggestion, though, if that is what you were asking.

okurz:
> > Can you comment why not what the ticket https://progress.opensuse.org/issues/165024 suggested to do as first alternative which is to do the retry within openqa-bootstrap?
> 
> I don't understand the question.
> 
> [#5840](https://github.com/os-autoinst/openQA/pull/5840) is about the second suggestion, though, if that is what you were asking.

Well, at the time of my review your second PR did not exist and I didn't know your intentions :) Now with the second PR present I can follow up to ask: Is this PR here actually needed given https://github.com/os-autoinst/openQA/pull/5840 ?

tests/install/openqa_webui.pm
```diff
@@ -79,7 +79,8 @@ sub install_containers {
 
 sub install_from_bootstrap {
     install_packages('openQA-bootstrap');
-    assert_script_run('skip_suse_specifics=1 skip_suse_tests=1 /usr/share/openqa/script/openqa-bootstrap', timeout => 1200);
+    # The maximum of the retry is 3810 seconds
+    assert_script_run(qq{retry -e -s 30 -r 7 -- sh -c "zypper -n --gpg-auto-import-keys ref && skip_suse_specifics=1 skip_suse_tests=1 /usr/share/openqa/script/openqa-bootstrap"}, 4000);
```
kalikiana, line 83: I feel like refreshing from within openqa-bootstrap is cleaner, and would lean towards https://github.com/os-autoinst/openQA/pull/5840

PR #197, opened 2024-08-13
Also upload system logs to find other errors

okurz:
Recently we observed a problem with the os-autoinst-openvswitch service.
While the relevant system journal content was shown in screenshots it
was not uploaded in an easy to process log file so I am adding another
call to also upload the complete system journal which should not be that
big on those clean test OS installations.

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -28,6 +28,7 @@ sub test_flags { {fatal => 1} }
 
 sub upload_openqa_logs {
     get_log 'journalctl --pager-end --no-tail --no-pager -u apache2 -u nginx -u openqa-scheduler -u openqa-websockets -u openqa-webui -u openqa-gru -u openqa-worker@1' => 'openqa_services.log.txt';
+    get_log 'journalctl --pager-end --no-tail --no-pager' => 'journal.log.txt';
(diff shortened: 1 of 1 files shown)


lib/openQAcoretest.pm
```diff
@@ -28,6 +28,7 @@ sub test_flags { {fatal => 1} }
 
 sub upload_openqa_logs {
     get_log 'journalctl --pager-end --no-tail --no-pager -u apache2 -u nginx -u openqa-scheduler -u openqa-websockets -u openqa-webui -u openqa-gru -u openqa-worker@1' => 'openqa_services.log.txt';
+    get_log 'journalctl --pager-end --no-tail --no-pager' => 'journal.log.txt';
```
kalikiana, line 31: Seems like overkill considering we already collect logs for services we care about - this is not "system" so much as everything including the "openQA services" 🤔

kalikiana: I'd say either rename it to **Always upload the whole journal** or filter it to be a subset outside of what we already save

okurz: you mean rename the git commit message?

kalikiana: Yes. Just make it clear it's the whole journal. No strong opinion otherwise.

okurz: ok, updated git commit message

perlpunk (approved):
I still think the sentence in the commit message is too long, but ok

Martchus (approved):

PR #203, opened 2024-08-29
Unify parsing test results

baierjan:
Take the best from both versions, show the output, allow better results parsing and remove one additional dependency (and potential point of failure).

Reference: https://progress.opensuse.org/issues/165692

### DIFF
--- a/tests/osautoinst/test_running.pm
+++ b/tests/osautoinst/test_running.pm
@@ -3,16 +3,18 @@ use testapi;
 use utils;
 
 sub run ($self) {
-    assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
+    my $api_query = get_var('FULL_MM_TEST') ? 'test=ping_client' : 'state=running state=done';
+    my $success = get_var('FULL_MM_TEST') ? 'passed' : 'passed\|running';
+    assert_script_run qq{retry -s 30 -r 30 -- sh -c '
+        r=`openqa-cli api jobs $api_query | tee /dev/fd/2 |
+        jq -r ".jobs | max_by(.id) | if .result != \\"none\\" then .result else .state end"`;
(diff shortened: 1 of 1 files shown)


tests/osautoinst/test_running.pm
```diff
@@ -3,17 +3,19 @@ use testapi;
...
 
 sub run ($self) {
-    assert_script_run 'command -v ack >/dev/null || zypper --no-refresh -n in ack';
+    my $api_query = get_var('FULL_MM_TEST') ? 'test=ping_client' : 'state=running state=done';
+    my $success = get_var('FULL_MM_TEST') ? 'passed' : 'passed\|running';
+    assert_script_run qq{retry -s 30 -r 30 -- sh -c '
+        r=`openqa-cli api jobs $api_query | tee /dev/fd/2 |
+        jq -r ".jobs | max_by(.id) | if .result != \\"none\\" then .result else .state end"`;
+        echo \$r | grep -q "incomplete\\|failed" && killall retry;
+        echo \$r | grep -q "$success"'}, timeout => 930;
```
perlpunk, line 12: We were using `ack` because it has the `--passthru` option to always the output even if it does not match.
This has helped a lot in the past, to see the actual content when it didn't match.

What is `tee /dev/fd/2` doing?

baierjan: > What is `tee /dev/fd/2` doing?

It shows the output even if it does not match. `stdin` is passed to `stdout` and copied to `stderr` (i.e. `/dev/fd/2`).

> We were using `ack` because it has the `--passthru` option to always the output even if it does not match. This has helped a lot in the past, to see the actual content when it didn't match.

Yes, that's what the tee is doing. See https://openqa.opensuse.org/tests/4443279#step/test_running/3

perlpunk (approved):

kalikiana (approved):

PR #207, opened 2024-09-07
Make test more stable by staying on the current page

baierjan:
Sometimes it take too long to load the new page and focus in the search bar can be lost upon page load. The search bar work on the current page as well so there is no need to visit all tests page.

### DIFF
--- a/tests/openQA/tests.pm
+++ b/tests/openQA/tests.pm
@@ -7,7 +7,6 @@ sub run {
     send_key_until_needlematch 'openqa-scheduled-test', 'down';
     assert_and_click 'openqa-scheduled-test';
     assert_screen 'openqa-test-details';
-    assert_and_click 'openqa-all-tests';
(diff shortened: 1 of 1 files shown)


Martchus (approved):

kalikiana (approved):

okurz:
But it's a good practice to leave the situation as it was in before so I would prefer if we end up on the "all tests" page again. Can we consider alternatives?

baierjan:
> But it's a good practice to leave the situation as it was in before so I would prefer if we end up on the "all tests" page again. Can we consider alternatives?

I agree, but the "all tests" page is not the correct page either. We should go back to the dashboard page, but that requires some additional needles for clicking so this was the fastest fix for the instability / sporadic issues we saw. We can of course improve and add the logic+needles in next steps. I can do that during next opportunity.

PR #208, opened 2024-09-12
Go back to dashboard page before next test module

baierjan:
This should target the concern from https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/207#issuecomment-2343002204

Needs needle from:
* https://github.com/os-autoinst/os-autoinst-needles-openQA/pull/29

### DIFF
--- a/tests/openQA/tests.pm
+++ b/tests/openQA/tests.pm
@@ -7,6 +7,9 @@ sub run {
     send_key_until_needlematch 'openqa-scheduled-test', 'down';
     assert_and_click 'openqa-scheduled-test';
     assert_screen 'openqa-test-details';
+    assert_and_click 'openqa-dashboard';
+    wait_still_screen;
+    assert_screen 'openqa-dashboard';
 }
 
 1;


Martchus (approved):

perlpunk (approved):

tests/openQA/tests.pm
```diff
@@ -7,6 +7,9 @@ sub run {
     send_key_until_needlematch 'openqa-scheduled-test', 'down';
     assert_and_click 'openqa-scheduled-test';
     assert_screen 'openqa-test-details';
+    assert_and_click 'openqa-dashboard';
+    wait_still_screen;
```
okurz, line 11: I would like to avoid all uses of `wait_still_screen` in "flagship" repositories, i.e. the ones that can be seen as good examples. Basically `wait_still_screen` is just a sneaky "sleep". As there is already a proper `assert_screen` afterwards: Why should this be necessary?

baierjan: Ok, that's doable but that would need an additional needle. The problem is that the openqa-dashboard needle used for clicking on the logo (to go to the dashboard page) can actually match on all screens. The instability we saw is because sometimes the needle is match before the new page is loaded. So I guess you prefer to have one needle for going to the dashboard and a second one to validate we are on that dashboard (which would be probably matching the welcome to openQA text / "jumbotron" area)?

okurz: Yes. It's better to have an explicit needle ensuring that the target page is sufficiently loaded and responsive again for any next test step. A separate needle is both more explicit about what the test is actually expecting as well as quicker as in most cases any needle check should really be much faster than the multi-second wait of `wait_still_screen`

baierjan: Ok, I will prepare yet another PR with that 😄

PR #210, opened 2024-09-16
Upload full openqa job testresults

b10n1k:
Compress and upload the testresults directory and the running processes for investigation.

https://progress.opensuse.org/issues/166445


tests/openQA/tests.pm
```diff
@@ -10,6 +10,8 @@ sub run {
     assert_and_click 'openqa-dashboard';
     wait_still_screen;
     assert_screen 'openqa-dashboard';
+    my @logs = split m/\n/, script_output q{find /var/lib/openqa -name autoinst-log.txt};
+    @logs and get_log "(cat @logs ||:)" => 'autoinst-log.txt';
```
Martchus, line 14: This could be simplified:
```suggestion
    get_log 'find /var/lib/openqa -name autoinst-log.txt -exec cat {} \+' => 'autoinst-log.txt';
```

perlpunk: the request in the original ticket was to only upload something if there is a logfile

okurz: and this is a copy-paste of the original code in lib/openQAcoretest. Please don't duplicate code

b10n1k: is it ok to get the autoinst-log.txt for all test results?
if so should I copy-paste the post-run-hook?
or none of them matter?

b10n1k: But lets me resolve this as I have removed it in the latest commits

perlpunk (approved):

Martchus (approved):

PR #211, opened 2024-09-20
Handle docker build failures

baierjan:
Upload docker logs when the build fails, make docker build logs available for auto review, detect and reference poo#165992 if applicable.

Reference: https://progress.opensuse.org/issues/165992


okurz (approved):
> WDYT

Overall this looks like a good improvement. I am just concerned that the instructions become more and more complicated hence don't serve as well anymore as a good example instructions that people could follow.

, or is the last commit too much?

It's good. You can leave it in.

PR #219, opened 2025-01-06
Unify log uploads

baierjan:
Logs in multi-machine scenario are uploaded twice, deduplicate the code and use the best from both in the main post-fail hook.

Reference: https://progress.opensuse.org/issues/169249

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -34,8 +34,9 @@ sub upload_openqa_logs {
     my @logs = split m/\n/, script_output q{find /var/lib/openqa -name autoinst-log.txt};
     @logs and get_log "(cat @logs ||:)" => 'autoinst-log.txt';
     get_log '(find /var/lib/openqa/pool/ /var/lib/openqa/testresults/ ||:)' => 'find.txt';
-    script_run 'tar -cvzhf testresults.tar.gz /var/lib/openqa/testresults/';
-    upload_logs 'testresults.tar.gz';
+    # do not assert for tar as it might return 1 due to the -h flag when dereferencing broken symlinks
+    script_run q{tar cahvf /tmp/testresults.tar.xz -C /var/lib/openqa/ testresults};
+    upload_logs "/tmp/testresults.tar.xz" if (script_run("test -f /tmp/testresults.tar.xz") == 0);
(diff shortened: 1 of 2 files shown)


tests/osautoinst/test_running.pm
```diff
@@ -13,7 +13,6 @@ sub run ($self) {
     if (get_var('FULL_MM_TEST')) {
         # we can't upload logs if the multimachine OVS bridge in the SUT has the same IP as the openQA-worker host
         script_run 'ip a del 10.0.2.2/15 dev br1'; # This may fail in case this IP is not actually set on the bridge
-        $self->upload_mm_logs();
```
perlpunk, line 16: There might have been a reason why this was called during the normal test and not only in the post_fail_hook.

baierjan: This is still true, look at the next line, `upload_openqa_logs` should handle that now as the log uploading was added to it recently.

lib/openQAcoretest.pm
```diff
@@ -34,8 +34,9 @@ sub upload_openqa_logs {
     my @logs = split m/\n/, script_output q{find /var/lib/openqa -name autoinst-log.txt};
     @logs and get_log "(cat @logs ||:)" => 'autoinst-log.txt';
     get_log '(find /var/lib/openqa/pool/ /var/lib/openqa/testresults/ ||:)' => 'find.txt';
-    script_run 'tar -cvzhf testresults.tar.gz /var/lib/openqa/testresults/';
-    upload_logs 'testresults.tar.gz';
+    # do not assert for tar as it might return 1 due to the -h flag when dereferencing broken symlinks
+    script_run q{tar czhvf /tmp/testresults.tar.gz -C /var/lib/openqa/ testresults};
+    upload_logs "/tmp/testresults.tar.gz" if (script_run("test -f /tmp/testresults.tar.gz") == 0);
```
okurz, line 39:
```suggestion
    script_run q{tar cahvf /tmp/testresults.tar.xz -C /var/lib/openqa/ testresults};
    upload_logs "/tmp/testresults.tar.xz" if (script_run("test -f /tmp/testresults.tar.gx") == 0);
```

okurz (changes requested):

lib/openQAcoretest.pm
baierjan: So we do prefer `xz` over `gzip`? I guess it should be just on this place so I can definitely easily change that.

okurz (approved):

perlpunk (approved):

PR #220, opened 2025-01-10
Visit both ping tests in multimachine scenario

baierjan:
Reference: https://progress.opensuse.org/issues/169249

Needs needles from:
* os-autoinst/os-autoinst-needles-openQA/pull/32

### DIFF
--- a/main.pm
+++ b/main.pm
@@ -63,7 +63,7 @@ sub load_shutdown() {
 load_install_tests();
 # testing from git only tests webui so far
 load_osautoinst_tests() unless get_var('OPENQA_FROM_GIT');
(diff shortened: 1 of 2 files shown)


Martchus:
The changes look good. It would be nice if you explained the reasoning why this change fixes the referenced issue.

baierjan:
> The changes look good. It would be nice if you explained the reasoning why this change fixes (or otherwise helps) the referenced issue.

It will probably not fix the issue, but it will (hopefully) make it more visible (as it is already the case for the simple boot test).

okurz (approved):

perlpunk (approved):

PR #229, opened 2025-03-13
Disable selinux as a temporary workaround

perlpunk:
Issue: https://progress.opensuse.org/issues/178822

### DIFF
--- a/tests/install/prepare.pm
+++ b/tests/install/prepare.pm
@@ -5,6 +5,9 @@ use utils qw(login disable_packagekit switch_to_root_console);
 sub run {
     login;
 
+    # Temorary workaround for selinux https://progress.opensuse.org/issues/178822
+    assert_script_run('semanage permissive -a httpd_t');
+
     disable_packagekit;
     assert_script_run('for i in {1..7}; do zypper --no-cd -n in retry && break; sleep $((i**2*20)); done');
     assert_script_run('zypper --no-cd -n rm xscreensaver');


tests/install/prepare.pm
```diff
@@ -5,6 +5,9 @@ use utils qw(login disable_packagekit switch_to_root_console);
 sub run {
     login;
 
+    # Temorary workaround for selinux https://progress.opensuse.org/issues/178822
```
nicksinger, line 8: maybe `record_soft_failure` to really remind us that this is only temporary?

nicksinger (approved):

tests/install/prepare.pm
kalikiana:
```suggestion
    # Temporary workaround for SELinux https://progress.opensuse.org/issues/178822
```

> maybe record_soft_failure to really remind us that this is only temporary?

It's not a failure at this point... 🤔

kalikiana (approved):

tests/install/prepare.pm
nicksinger: true, never mind then

okurz: @nicksinger I think your suggestion is valid and it seems @kalikiana might not know what we use `record_soft_failure` for usually. Here for sure we could use `record_soft_failure` to record that reference and IMHO it would be helpful however we should then combine it with a check to see if the workaround is actually needed. Here it's applied unconditionally.

kalikiana: Right. I was missing the words. The condition that will eventually make it work is what's missing here.

PR #230, opened 2025-03-13
Update package installation commands with ZYPP_PCK_PRELOAD

b10n1k:


okurz:
Why?

b10n1k:
> Why?

it is quite faster. I guess that will be enabled by default in the future and we cant revert these changes.

kalikiana:
> > Why?
> 
> it is quite faster. I guess that will be enabled by default in the future and we cant revert these changes.

I would suggest mentioning your motivation in the description. Everyone may not have read the announcement on [factory@lists.opensuse.org](https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/LOCZIG43MFJSTUIQ3VH2CRSYRCBNR4O7/). That said, you have my vote.

Martchus (changes requested):
The commit messages has grammar mistakes and most importantly states that this enable parallel installation (but in reality this change enables parallel downloads).

Martchus:
I see that you've just updated the commit but the commit message still says "parallel installation". This is wrong. The feature enable here will only enable parallel downloads. And "We might expect some that it works …" still sounds wrong.

I would word it this way:

```
Enable parallel downloads during package installation

Use ZYPP_PCK_PRELOAD=1 to enable parallel downloads during package installations
to make it faster.
```

PR #232, opened 2025-03-27
Bump retry timeout in test_running to 90s

kalikiana:
Whilst this only fails occasionally the easiest option is to increase the timeout.

See: https://progress.opensuse.org/issues/179512


tests/osautoinst/test_running.pm
```diff
@@ -5,7 +5,7 @@ use utils;
 sub run ($self) {
     my $api_query = get_var('FULL_MM_TEST') ? 'test=ping_client' : 'state=running state=done';
     my $success = get_var('FULL_MM_TEST') ? 'passed' : 'passed\|running';
-    assert_script_run qq{retry -s 30 -r 30 -- sh -c '
+    assert_script_run qq{retry -s 90 -r 30 -- sh -c '
```
okurz, line 8: I assume in the best case the call returns with success much sooner than 30s so I suggest to keep the lower sleep period. I suggest to just do more retries

kalikiana: It takes too long to execute even if we try 30 times. That's also what we discussed during the estimation.

Do you think 300 retries would eventually succeed?

kalikiana: Increase the number of seconds we sleep
We seem to be exhausting the 30 seconds wait setting up the test before we get to cloning

okurz: Well, the problem was that depending on the test setup and configuration it can be that osado is cloned in full history which can just take a long time as osado is a huge git repo. Hence we need to wait long enough here overall. Just do 90 retries meaning that we wait in worst case 30s*90=2700s=45m. That's the same time as with 90s*30 but we check more often if the expected result might be already there so saving time in the best case.

perlpunk: it's very unlikely IMHO that the inner test is finished already after less than 30s, if we consider git cloning plus the runtime of the inner test itself.
I would just double the sleep time to 60. I think in all the cases I looked at that would have been more than enough time.

okurz: No I don't expect the inner test to usually complete within 30s but 30s is effectively the checking interval. If we bump to 90s that means that in the worst case the test finishes at N*90s+1s but we waste another 89s until we even look hence my suggestion to keep the 30s. We could actually go even lower to check more often. It's just an API call to localhost

perlpunk: well, I don't care much.
in any case, I would only double the number of retries or seconds, not triple.

kalikiana: Will make it 15 and 120 then as discussed

perlpunk (approved):

tests/osautoinst/test_running.pm
```diff
@@ -5,7 +5,7 @@ use utils;
 sub run ($self) {
     my $api_query = get_var('FULL_MM_TEST') ? 'test=ping_client' : 'state=running state=done';
     my $success = get_var('FULL_MM_TEST') ? 'passed' : 'passed\|running';
-    assert_script_run qq{retry -s 30 -r 30 -- sh -c '
+    assert_script_run qq{retry -s 15 -r 120 -- sh -c '
```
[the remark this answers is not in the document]
okurz: Now it's *exactly* the same value inviting for flaky behavior. That's why in before it was 930, not 900. How about 1830?

perlpunk (approved):

okurz (approved):

PR #233, opened 2025-03-27
Avoid using the redirector service with devel:openQA

kalikiana:
Otherwise packages being added and removed can lead to errors while fetching packages from the repo.

See: https://progress.opensuse.org/issues/178642

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -12,7 +12,9 @@ sub add_repo {
     );
     my $repo = 'openSUSE_' . $repo_suffix{get_required_var('ARCH')};
     my $project = get_var('OPENQA_OBS_PROJECT', 'devel:openQA');
-    my $repo_url = get_var('OPENQA_REPO_URL', "obs://$project/$repo");
+
+    # Avoid using the redirector service to cope with transient repo content
+    my $repo_url = get_var('OPENQA_REPO_URL', "https://downloadcontent.opensuse.org/repositories/$project/$repo");
     assert_script_run("zypper -n ar -p 95 -f '$repo_url' openQA");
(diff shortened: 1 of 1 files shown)


okurz:
Does it work?

kalikiana:
> Does it work?

I think I misunderstood the semantics of PEDANTIC. Apparently it will ensure files are mirrored, but also fail if the packages are gone.

So that leaves the second option, which is using downloadcontent.opensuse.org.

tests/install/openqa_webui.pm
```diff
@@ -12,7 +12,9 @@ sub add_repo {
     );
     my $repo = 'openSUSE_' . $repo_suffix{get_required_var('ARCH')};
     my $project = get_var('OPENQA_OBS_PROJECT', 'devel:openQA');
-    my $repo_url = get_var('OPENQA_REPO_URL', "obs://$project/$repo");
+
+    # Using PEDANTIC here to skip the redirector service
```
okurz, line 16: Now this comment does not apply anymore. Please update it to also reflect that this only done as an exceptional case as we work with a transient repository content.

okurz (changes requested):

kalikiana:
> So ... does it work?

I don't know what else I am supposed to say. If the reasoning I provided is not good enough I'll probably put the ticket up for re-estimation

okurz:
> > So ... does it work?
> 
> I don't know what else I am supposed to say. If the reasoning I provided is not good enough I'll probably put the ticket up for re-estimation

There is nothing to re-estimate. I was just asking for verification. So how did you test?

kalikiana:
> There is nothing to re-estimate. I was just asking for verification. So how did you test?

By running the job and inspecting the results. I didn't make any changes to production.

PR #234, opened 2025-04-04
Fix updating authentication method with recent ini-file changes

okurz:
CI tests fail needing https://github.com/os-autoinst/openQA/pull/6363

Related progress issue: https://progress.opensuse.org/issues/179359

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -36,7 +36,7 @@ sub install_from_pkgs {
         assert_script_run 'systemctl restart apache2';
     }
     assert_script_run($_) foreach (split /\n/, <<~'EOF');
-    sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini
+    if [ -e /etc/openqa/openqa.ini ]; then sed -i -e 's/#.*method.*OpenID.*$/&\nmethod = Fake/' /etc/openqa/openqa.ini; else echo -e "[auth]\nmethod = Fake" > /etc/openqa/openqa.ini.d/auth.ini; fi
     systemctl enable --now openqa-webui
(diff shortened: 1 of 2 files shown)


Martchus (approved):

Martchus (commented):
Why do we need the `\` only after the first line?

okurz:
> Why do we need the `\` only after the first line?

Well, we don't need it. See https://openqa.opensuse.org/tests/4970357#step/openqa_webui/13 :) I guess I should join up the line after `then`

Martchus (approved):

b10n1k (approved):
approved but it would be nice to have more in the commit message than the poo link

okurz:
> approved but it would be nice to have more in the commit message than the poo link

ok. I put more details into both git commits

perlpunk (approved):

okurz:
@Martchus can you please look into https://openqa.opensuse.org/tests/4977191#step/start_test/5 showing "permission denied"? https://openqa.opensuse.org/tests/4977191#step/openqa_webui/5 shows that we run ff535184 which includes https://github.com/os-autoinst/openQA/pull/6366 but something seems to be still missing

Martchus:
Strange, I updated the apparmor profile for the client in ff535184 as well.

Martchus:
I cloned this test and entered the VM with the developer mode. In a session as user bernhard `openqa-cli schedule` generally works (I haven't tried with all parameters). When using sudo it works as well. I guess the problem here is actually not `opena-cli` but the server itself as we actually get an error message *from* the server. I'm wondering why the server is using the client config here, though.

kalikiana:
@mergifyio rebase

okurz:
openQA-in-openQA  tests now fail in a later step due to https://progress.opensuse.org/issues/180002

PR #237, opened 2025-04-10
Disable git_clone minion jobs

perlpunk:
We are using fetchneedles with --depth 1 for the initial clone. We don't need to fetch the whole osado repo in this test.

As a result of this we can also reduce the retry attempts and timeout again.

After #236


perlpunk:
> Or as alternative why do we even still call fetchneedles? I guess https://progress.opensuse.org/issues/164898 at least from the _title_ point of view was not yet completed?

From https://progress.opensuse.org/issues/164898
> Out of scope
> Doing any kind of initial checkout if git working copies do not exist yet

okurz:
> > Or as alternative why do we even still call fetchneedles? I guess https://progress.opensuse.org/issues/164898 at least from the _title_ point of view was not yet completed?
> 
> From https://progress.opensuse.org/issues/164898
> 
> > Out of scope
> > Doing any kind of initial checkout if git working copies do not exist yet

true true. wasn't that done nevertheless as part of another ticket then?

perlpunk:
> > From https://progress.opensuse.org/issues/164898
> > > Out of scope
> > > Doing any kind of initial checkout if git working copies do not exist yet
> 
> true true. wasn't that done nevertheless as part of another ticket then?

Nope. At least not for jobs that don't have CASEDIR/NEEDLES_DIR

perlpunk:
> Wouldn't it be a good test for real behaviour to keep the git_clone minion jobs?

openqa_install+publish and openqa_from_bootstrap are still using that for running a distri-example test

tests/install/openqa_webui.pm
```diff
@@ -35,6 +35,10 @@ sub install_from_pkgs {
         assert_script_run 'systemctl enable apache2';
         assert_script_run 'systemctl restart apache2';
     }
+    if (get_var('FULL_OPENSUSE_TEST')) {
+        # avoid second git fetch after fetchneedles
+        assert_script_run q{if [ -e /etc/openqa/openqa.ini ]; then sed -i -e 's/#\[scm git\]/[scm git]\ngit_auto_clone = no\ngit_auto_update = no/' /etc/openqa/openqa.ini; else echo -e "[scm git]\ngit_auto_clone = no\ngit_auto_update = no" > /etc/openqa/openqa.ini.d/git.ini; fi};
```
Martchus, line 40: Maybe we can just assume that `/etc/openqa/openqa.ini.d` is supported.

perlpunk: > Why is this in the `install_from_pkgs`.

That's the only place where we edit openqa.ini currently. Directly below this.

perlpunk: > Maybe we can just assume that `/etc/openqa/openqa.ini.d` is supported.

I just used the same as below, so I'd say it would just look weird if we still have that logic below but not in the code I introduce.

b10n1k: Dont want to be destructive but I think it make sense to take those openqa.ini out of that function, wrap them into another with a corresponding name and add it in the `run` function. it will look more clear and will keep the functions clean. wdyt?

perlpunk: I'm not sure why you're all complaining about this line but did not complain about line 43 below

Martchus: Because line 43 was introduced when the new config parsing was just introduced so it had to support the old and new behavior. However, I don't think this is required anymore so always creating a drop-in config file should be enough.

PR #238, opened 2025-04-15
Fix error 403 when accessing dashboard & re-enable Nginx

realcharmer:
https://progress.opensuse.org/issues/180002#note-19

Verification runs:
- Apache2: https://openqa.opensuse.org/tests/4995318
- Nginx: https://openqa.opensuse.org/tests/4995439


Martchus (approved):
So we even have documented this: https://open.qa/docs/#_nginx_proxy

I guess then we can merge this PR. Not sure whether it would make sense to change the config template nevertheless.

realcharmer:
> So we even have documented this: https://open.qa/docs/#_nginx_proxy
 
Indeed, but I think it could be worded better or even include an example for the edit. But since it's documented, changing the url from `localhost` to `127.0.0.1` is a better option compared to editing the Nginx configuration.

Also I am not sure, why it would start failing recently, because it was running fine before.

perlpunk (approved):

PR #240, opened 2025-05-02
Fix dashboard access on nginx

realcharmer:
https://progress.opensuse.org/issues/180002

Change the default nginx configuration so the dashboard can be accessed from `localhost` as well as an IP address. The test coverage is also expanded to include all options for accessing the dashboard.

Verification run: https://openqa.opensuse.org/tests/5035573

### DIFF
--- a/tests/install/openqa_webui.pm
+++ b/tests/install/openqa_webui.pm
@@ -28,6 +28,7 @@ sub install_from_pkgs {
     assert_script_run "/usr/share/openqa/script/configure-web-proxy $proxy_args";
     if (check_var('OPENQA_WEB_PROXY', 'nginx')) {
         assert_script_run 'systemctl disable --now apache2';
+        assert_script_run 'sed -i "/server_name/ s/localhost/_/" /etc/nginx/nginx.conf';
         assert_script_run 'systemctl enable nginx';
         assert_script_run 'systemctl restart nginx';
     }
--- a/tests/openQA/dashboard.pm
+++ b/tests/openQA/dashboard.pm
@@ -8,8 +8,11 @@ sub run {
     prepare_firefox_autoconfig;
     switch_to_x11;
     ensure_unlocked_desktop();
+    start_gui_program('firefox http://localhost', 60, valid => 1);
+    assert_screen 'openqa-dashboard', 60;
     start_gui_program('firefox http://127.0.0.1', 60, valid => 1);
-    #wait few minutes for ff to start and then fail the test
-    assert_screen 'openqa-dashboard', 600;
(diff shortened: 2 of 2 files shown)


tests/openQA/dashboard.pm
```diff
@@ -8,8 +8,11 @@ sub run {
     prepare_firefox_autoconfig;
     switch_to_x11;
     ensure_unlocked_desktop();
+    start_gui_program('firefox http://localhost', 60, valid => 1);
+    assert_screen 'openqa-dashboard', 60;
```
realcharmer, line 12: Are the delays here fine? I reckon this could be optimized further.

kalikiana: Note that this is 60s max in the worst case, not a delay.

okurz: I suggest to not even do those additional checks here by starting firefox as that is rather expensive. Just revert the code to check for localhost. If that works then we can be sure that also 127.0.0.1 and [::1] works. We should focus the openQA-in-openQA tests to test usual use cases, not cover all corner cases.

tests/install/openqa_webui.pm
```diff
@@ -28,6 +28,7 @@ sub install_from_pkgs {
     assert_script_run "/usr/share/openqa/script/configure-web-proxy $proxy_args";
     if (check_var('OPENQA_WEB_PROXY', 'nginx')) {
         assert_script_run 'systemctl disable --now apache2';
+        assert_script_run 'sed -i "/server_name/ s/localhost/_/" /etc/nginx/nginx.conf';
```
okurz, line 31: I guess this can go into configure-web-proxy, see L28

Shouldn't https://github.com/os-autoinst/openQA/blob/master/script/configure-web-proxy#L49 already do that? Maybe just this regex broke? @asdil12 introduced that initially with https://github.com/os-autoinst/openQA/commit/4fd97ce7237a45bbd83af7d874f54447903fbc38#diff-971564c2eee0f87443db77101fd84c3eb432ce45547d44178d1f1cae179ab83aR24-R33

okurz (changes requested):

realcharmer:
Alright, I took the wrong approach. I'll close this and create new PR after I fix the core issue. Thanks for the suggestions.

b10n1k:
I run the VR and checked the audit.log, which still shows `denial`. shouldnt this fix the selinux permissions?

okurz:
> I run the VR and checked the audit.log, which still shows `denial`. shouldnt this fix the selinux permissions?

No. This has nothing to do with SELinux

PR #246, opened 2025-07-30
WIP: Adjust test modules to run against SLE

b10n1k:
The test was adjusted using openqa_install+publish from O3.

https://progress.opensuse.org/issues/127286


okurz:
Please share verification jobs

lib/utils.pm
```diff
@@ -57,6 +64,26 @@ sub wait_for_desktop {
     assert_screen 'generic-desktop';
 }
 
+# sle
```
okurz, line 67: I think you are mixing up things. What you add is something like a completely new module to support initial setup of a JeOS image, not "SLE" per se. As mentioned in the related needles PR please just use the same base as equivalent on openSUSE, i.e. a "create_hdd_minimalx" job

okurz (changes requested):

PR #247, opened 2025-08-04
Validate conf files in the worker container

d3flex:
Just a small improvement in order to validate what is mount in the container as the `wait_for_container_log` confirms that it cannot find any keys but we do not know whether the configs which reads are as expected.

related poo: https://progress.opensuse.org/issues/186651

### DIFF
--- a/tests/containers/worker.pm
+++ b/tests/containers/worker.pm
@@ -3,9 +3,31 @@ use testapi;
 use utils;
 
 sub run {
+  my %confs = (
+      client_conf => {
+          path => 'conf/client.conf',
+          output => ''
+      },
+      worker_ini => {
+          path => 'conf/workers.ini',
+          output => ''
+      }
+);
   my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/openqa_data/data.template/conf/:/data/conf:ro"';
+  my $data_container_path = '/root/openQA/container/openqa_data/data.template/';
+  $confs{client_conf}{output} = script_output("cat $data_container_path$confs{client_conf}{path}");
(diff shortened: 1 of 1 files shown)


tests/containers/worker.pm
```diff
@@ -3,9 +3,31 @@ use testapi;
...
+          output => ''
+      }
+);
   my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/openqa_data/data.template/conf/:/data/conf:ro"';
+  my $data_container_path = '/root/openQA/container/openqa_data/data.template/';
+  $confs{client_conf}{output} = script_output("cat $data_container_path$confs{client_conf}{path}");
+  $confs{worker_ini}{output} = script_output("cat $data_container_path$confs{worker_ini}{path}");
+
   assert_script_run("docker run -d --network testing $volumes --name openqa_worker openqa_worker");
   wait_for_container_log('openqa_worker', 'API key and secret are needed', 'docker');
```
kalikiana, line 22: Why aren't you addressing this error message? If that's gone we would know config was ready and the connection works, no?

d3flex:
> so you're basically just checking whether the volume option in docker works correctly?

Ensure that the `API key and secret are needed` is indeed because the absence of values for the keys.
At first, we thought that the problem was with the client.conf. I think with this, it will be more visible what is expected.

tests/containers/worker.pm
kalikiana: Right now the **worker** test prepares a container and validates that it is not able to connect to a web UI.

If you're saying we should test both this, and the case of a successful connection I would say, yes, that makes sense.

d3flex: So we want to redesign the modules?
Does it mean that it will be better a multimachine test?

in any case this PR is just an additional check. it wasnt intented to change how things are now.

PR #248, opened 2025-08-08
Prevent worker to fail on missing key and secret keys

d3flex:
The worker still fails but not due to misconfuration. But this time will throw an additional soft fail to make visible the issue with the containers' communication.

poo: https://progress.opensuse.org/issues/186651

### DIFF
--- a/tests/containers/single_container_webui.pm
+++ b/tests/containers/single_container_webui.pm
@@ -12,7 +12,7 @@ sub run {
   wait_for_container_log('openqa_webui', 'Web application available at', 'docker');
 
   assert_script_run('curl http://localhost');
-  assert_script_run('docker rm -f openqa_webui');
+  #assert_script_run('docker rm -f openqa_webui');
 }
 
 sub post_fail_hook {
--- a/tests/containers/worker.pm
+++ b/tests/containers/worker.pm
@@ -3,10 +3,40 @@ use testapi;
 use utils;
 
 sub run {
-  my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/openqa_data/data.template/conf/:/data/conf:ro"';
-  assert_script_run("docker run -d --network testing $volumes --name openqa_worker openqa_worker");
-  wait_for_container_log('openqa_worker', 'API key and secret are needed', 'docker');
-  clear_root_console;
+    #my $apikey = '1234567890ABCDEF';
+    #my $apikey = '1234567890ABCDEF';
+    my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/openqa_data/data.template/conf/:/data/conf:ro"';
+    script_output(
+        "echo  \"\$(cat <<EOF
+[openqa_webui]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+
+[localhost]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+EOF
+        )\"  > /root/openQA/container/openqa_data/data.template/conf/client.conf"
+	);
+    script_output(
+        "echo  \"\$(cat <<EOF
+[global]
+BACKEND = qemu
+HOST = http://openqa_webui
+WORKER_HOSTNAME = openqa_worker
+EOF
+        )\"  > /root/openQA/container/openqa_data/data.template/conf/workers.ini"
+	);
+    assert_script_run("docker run -d --network testing $volumes --name openqa_worker openqa_worker");
+    my $expected_log = 'Failed to register';
+    wait_for_container_log('openqa_worker', 'Failed to register', 'docker');
+    record_soft_failure("$expected_log - https://progress.opensuse.org/issues/186651");
+    clear_root_console;
+}
+
+sub post_run_hook {
+  script_run('docker rm -f openqa_worker');
+  script_run('docker rm -f openqa_webui');
 }
 
 1;


tests/containers/worker.pm
```diff
@@ -3,10 +3,40 @@ use testapi;
...
+BACKEND = qemu
+HOST = http://openqa_webui
+WORKER_HOSTNAME = openqa_worker
+EOF
+        )\"  > /root/openQA/container/openqa_data/data.template/conf/workers.ini"
+	);
+    assert_script_run("docker run -d --network testing $volumes --name openqa_worker openqa_worker");
+    my $expected_log = 'Failed to register';
+    wait_for_container_log('openqa_worker', 'Failed to register', 'docker');
+    record_soft_failure("$expected_log - https://progress.opensuse.org/issues/186651");
```
kalikiana, line 33: Why do we need a soft fail? If you're adding valid API credentials to the config, this should work.

okurz: And please only use softfail references with product issues, not progress tickets.

d3flex: > Why do we need a soft fail? If you're adding valid API credentials to the config, this should work.

The idea was to let a green test, but somehow make visible that this is not functional, so it wont be ignored. As it was the case so far and surprised us when we found out.
The problem is not about the keys anymore. I didnt look at it since I pushed the draft, but i think it is related to the containers issue and the proxy setup or something. please try out and I would be glad to sit down together on Monday if you want.

> And please only use softfail references with product issues, not progress tickets.

I dont know if you have the context. the test is passing while the worker module is never connecting to the server. For various reasons but also by design from my POV. I am looking for ideas. would just a record_info enough? the test will still shown as passed. not sure if I am clear or I need to say more

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,54 @@ use testapi;
 use utils;
 
 sub run {
+  script_output(
```
okurz, line 6: why script_output when you don't use the output?

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,54 @@ use testapi;
...
   my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/webui/conf:/data/conf:ro"';
   my $certificates = '-v "/root/server.crt:/etc/apache2/ssl.crt/server.crt" -v "/root/server.crt:/etc/apache2/ssl.crt/ca.crt" -v "/root/server.key:/etc/apache2/ssl.key/server.key"';
 
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --hostname openqa_webui --name openqa_webui openqa_webui");
   wait_for_container_log('openqa_webui', 'Web application available at', 'docker');
-
+
```
okurz, line 49: ?

d3flex:
> Nice. I didn't know that we run apache in the webui container.
> 
> Maybe the `curl http://openqa_webui/login` can be done as part of the webui startup, then the entrypoint of the worker can be used as it is.

I should give it a go. But I presume there are nuances for this scenario?! what happens if the authentication is not using `Fake`?

tests/containers/single_container_webui.pm
d3flex: ignore

tests/containers/worker.pm
```diff
@@ -3,10 +3,35 @@ use testapi;
...
-  wait_for_container_log('openqa_worker', 'API key and secret are needed', 'docker');
-  clear_root_console;
+    my $volumes = '-v "/root/data/factory:/data/factory" -v "/root/data/tests:/data/tests" -v "/root/openQA/container/openqa_data/data.template/conf/:/data/conf:ro"';
+    script_output(
+        "echo  \"\$(cat <<EOF
+[openqa_webui]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+
+[localhost]
```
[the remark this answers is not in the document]
d3flex: The question is do you think that it has to have the minimum for this case?

okurz:
> > Nice. I didn't know that we run apache in the webui container.
> > Maybe the `curl http://openqa_webui/login` can be done as part of the webui startup, then the entrypoint of the worker can be used as it is.
> 
> I should give it a go. But I presume there are nuances for this scenario?! what happens if the authentication is not using `Fake`?

I suggest to do what's necessary for the test to work in a clean and simple way so it's ok if the same approach wouldn't work in other scenarios

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,54 @@ use testapi;
...
+secret = 1234567890ABCDEF
+
+[scheduler]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF
+
+[websockets]
+key = 1234567890ABCDEF
+secret = 1234567890ABCDEF    
+[openqa_webui]
```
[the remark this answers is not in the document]
perlpunk: Then using append mode `>>` might be the better choice here?

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,39 @@ use testapi;
...
 
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --hostname openqa_webui --name openqa_webui openqa_webui");
   wait_for_container_log('openqa_webui', 'Web application available at', 'docker');
-
-  assert_script_run('curl http://localhost');
-  assert_script_run('docker rm -f openqa_webui');
+  assert_script_run('curl -v http://localhost/login');
```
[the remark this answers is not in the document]
kalikiana: > However I wonder why we need to duplicate what's already in https://github.com/os-autoinst/openQA/blob/e6049dfe05201ff1bb9cf8333b62b2fc12747f0e/script/openqa-bootstrap#L176 .

We are not duplicating anything. FakeAuth requires a call to `login` to create a user. [openqa-bootstrap](https://github.com/os-autoinst/openQA/blob/master/script/openqa-bootstrap) sets up web UI, worker and other services - we would have to modify it to support only setting up a web UI. Currently we only use it in [the single instance container](https://github.com/os-autoinst/openQA/blob/master/container/single-instance/Dockerfile) where that is what we want.

tests/containers/single_container_webui.pm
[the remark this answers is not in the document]
kalikiana: As discussed:
* Please don't create a client.conf here - this should be done in the worker container.
* Also drop the openid & logging sections here which are not needed.

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,39 @@ use testapi;
...
+)\" > /root/openQA/container/webui/conf/client.conf");
+  script_run(
+        "echo  \"\$(cat <<EOF
+[global]
+# change to the URL the web UI will be available under so redirection for
+# authentication works
+base_url = http://openqa_webui
+
+[auth]
+method = Fake
```
okurz, line 21: https://github.com/os-autoinst/openQA/pull/6766 should help you to not need to put this into the config file.

tests/containers/single_container_webui.pm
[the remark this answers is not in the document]
okurz: As stated before move this to the worker.pm test module

tests/containers/worker.pm
```diff
@@ -3,10 +3,34 @@ use testapi;
...
+)\" > $confdir/client.conf");
+
+    script_run(
+        "echo  \"\$(cat <<EOF
+[global]
+BACKEND = qemu
+HOST = openqa_webui
+WORKER_HOSTNAME = openqa_worker
+EOF
+)\" > $confdir/workers.ini");
```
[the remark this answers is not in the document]
d3flex: I havent managed to make this work. it makes the process hanging, so I will leave it out for now

```# Test died: command 'cat > /tmp/openqa_worker_conf/workers.ini <<EOF
[global]
BACKEND = qemu
HOST = openqa_webui
WORKER_HOSTNAME = openqa_worker
EOF
' timed out at /usr/lib/os-autoinst/autotest.pm line 515.```

File: tests/containers/single_container_webui.pm [hunk 9]
```diff
@@ -3,16 +3,26 @@ use testapi;
...
 
   assert_script_run("openssl req -newkey rsa:4096 -x509 -sha256 -days 365 -nodes -subj '/CN=www.mydom.com/O=My Company Name LTD./C=DE' -out server.crt -keyout server.key");
 
-  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --name openqa_webui openqa_webui");
+  assert_script_run("docker run -d --network testing $volumes $certificates -p 80:80 --hostname openqa_webui --name openqa_webui openqa_webui");
   wait_for_container_log('openqa_webui', 'Web application available at', 'docker');
-
-  assert_script_run('curl http://localhost');
-  assert_script_run('docker rm -f openqa_webui');
+  assert_script_run('curl -v http://localhost');
```
okurz, line 25: Why the -v?

tests/containers/single_container_webui.pm
```diff
@@ -3,16 +3,26 @@ use testapi;
...
+        "echo  \"\$(cat <<EOF
+[global]
+# change to the URL the web UI will be available under so redirection for
+# authentication works
+base_url = http://openqa_webui
+
+[auth]
+method = Fake
+EOF
+)\" > /root/openQA/container/webui/conf/openqa.ini");
```
okurz, line 16: I suggest to use an override file and not the complete config openqa.ini


```suggestion
)\" > /root/openQA/container/webui/conf/openqa.ini.d/fake-auth.ini");
```

d3flex: But the container doesnt provide this directory. I will have to create it manually. Altrenative , is to modify the Dockerfile again. Which I would say lets do it in another PR/ticket

tests/containers/single_container_webui.pm
d3flex: leftover. removed

tests/containers/worker.pm
```diff
@@ -3,10 +3,34 @@ use testapi;
...
+    
+    assert_script_run(qq{docker run -d --network testing $volumes --hostname openqa_worker --name openqa_worker openqa_worker});
+    wait_for_container_log('openqa_worker', 'Registered and connected', 'docker');
+    clear_root_console;
+}
+
+sub post_run_hook {
+  script_run('docker rm -f openqa_worker');
+  script_run('docker rm -f openqa_webui');
 }
```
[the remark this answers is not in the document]
kalikiana: The `openqa_webui` container is no longer being removed in `tests/containers/single_container_webui.pm` . That's why this is needed now.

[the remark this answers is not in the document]
kalikiana: If we wanted to avoid this dependency, we would need to setup & run a web UI container within the worker test. That was discussed before.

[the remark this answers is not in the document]
okurz: I think originally the webUI container was removed to not influence any other test modules after that which is a good approach for testing. You now wanted to change the test from a "component test" to an "integration test" so the worker module relies on the webui test module. Fine for me. I would have not cared about removing the containers in this module then because the whole test flow became an integration test and left cleanup to the whole run, i.e. not call `docker rm` as afterwards the VM is destroyed anyway

tests/containers/single_container_webui.pm
nicksinger: How is this MR supposed to help here? What would be the value to set `-m` to? Reading https://open.qa/docs/#_fake it doesn't mention any other method to enable fake authentication. Or is something like https://github.com/os-autoinst/openQA/blob/05b96119875a37729e11d143ebbc6a402c96d50c/lib/OpenQA/Schema.pm#L39 also implemented in the auth-backend?

tests/containers/worker.pm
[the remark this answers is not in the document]
perlpunk: Ok, it doesn't work because `assert_script_run` appends the` ; echo ~hel1-\$?- ` stuff, but that doesn't work on its own line after the `EOF` line.
However, this works:
```perl
    type_string(<<~"EOSCRIPT");
    cat > $confdir/workers.ini <<EOF
    [global]
    BACKEND = qemu
    HOST = openqa_webui
    WORKER_HOSTNAME = openqa_worker
    EOF
    EOSCRIPT
```
and we already do this kind of command in `tests/install/openqa_worker.pm` for example.

I find this more readable, but I don't insist on it.

tests/containers/worker.pm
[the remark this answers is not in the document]
okurz: Better just remove the complete `post_run_hook`. We don't need to make the test run longer to cleanup services as the whole VM will be destroyed anyway

tests/containers/single_container_webui.pm
okurz: By setting the mode to "development" or "test" the fake authentication *and* log=debug is automatically enabled in https://github.com/os-autoinst/openQA/blob/05b96119875a37729e11d143ebbc6a402c96d50c/lib/OpenQA/Setup.pm#L270 . So instead of writing a config file one just needs to set the environment variable `OPENQA_WEBUI_MODE=test` when starting the openQA web UI container

PR #252, opened 2025-09-04
Simplify workflow

perlpunk:
Default values can already be set in the `env` block.

Related issue: https://progress.opensuse.org/issues/188310


perlpunk:
And it's failing currently because of https://progress.opensuse.org/issues/188310

kalikiana (approved):
Nice

okurz (approved):

PR #276, opened 2026-03-30
Explore worker confinement adding dedicate module

d3flex:
Provide a set of scripts which explore the confined capabilities of the workers. There are two main approaches: using the test API to read/write from within the SUT and the Host level operations using system().

RFD: https://github.com/os-autoinst/openQA/pull/7127

Issue: https://progress.opensuse.org/issues/194717

### DIFF
--- a/tests/openQA/worker_sec_checks.pm
+++ b/tests/openQA/worker_sec_checks.pm
@@ -1,0 +1,154 @@
+use Mojo::Base 'openQAcoretest', -signatures;
+use Mojo::File qw(path);
+use testapi;
+
+# Summary: Run and Verify workers' confinement
+# First stage runs operations on the SUT
+# - Read the shadow password file
+# - Read sudoers configuration
+# - exposes secrets passed at boot
+# - various operations
+#
+# Second stage runs host-level operations
+# - write to host /etc.
+# - write to the worker's own home directory.
+# - verify the cache directory is readable. ReadOnlyPaths must include it for the cache service to work
+# - try to write to the cache directory. should be blocked when ReadOnlyPaths is active
+# - verify this worker slot cannot read another slot's pool directory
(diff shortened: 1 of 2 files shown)


okurz (changes requested):

Martchus (commented):
I find it interesting that you've chosen to explore this as an openQA tests. This surely seems like a lot of work just for *exploring*.

Note that it is not forbidden to use signatures, `Mojo::File` and other "modern" features/libraries just because it is test code.

Is autodie enabled in test modules? If not you should do error handling after calls like `system` and `open` consistently. If autodie is enabled you should avoid explicit error handling consistently.

tests/openQA/worker_sec_checks.pm
```diff
@@ -0,0 +1,155 @@
...
+# verify the cache directory is readable — ReadWritePaths must include it for the cache service to work
+# try to write to the cache directory — should be blocked when ReadOnlyPaths is active
+# verify this worker slot cannot read another slot's pool directory
+# can we read a sensitive host file from Perl?
+# can we read another process /proc/<pid>/environ on the host? 
+
+sub run ($self) {
+    # select_console 'root-console';
+    send_key('ctrl-alt-f3');
+    record_info('VM-level', 'exercise API commands inside Vm');
```
Martchus:
```suggestion
    record_info('VM-level', 'exercise API commands inside VM');
```

Martchus (commented):
I guess this is generally useful. If we proceed with the confinement tickets we can also schedule such a test job to test other ways of confinement.

The upfront comments could a bit more concise and should include an example command for how to schedule a job using this test module.

You could also split the run function and the length upfront comment so you end up with e.g. `_test_sut_operations` and `_test_host_operations` that both have their individual pod documentation.

PR #280, opened 2026-04-01
Fix commit message checker workflow

perlpunk:
See commits

### DIFF
--- a/.github/workflows/commit_message_checker.yml
+++ b/.github/workflows/commit_message_checker.yml
@@ -10,4 +10,6 @@ on:
 
 jobs:
   check-commit-message:
+    permissions:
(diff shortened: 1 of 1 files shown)


okurz (approved):

d3flex (commented):
Small typo in `fix: Adapt commit_message_checker.yml to latest version`. the _input_ is written as `inut`

d3flex (commented):
I looked up at https://docs.github.com/en/actions/concepts/security/github_token and other mentioned links. where does it say `accessToken which isn't accepted as inut in the new workflow`. I am pretty sure that your changes are fine but just for clarity

Martchus (approved):

PR #284, opened 2026-04-16
fix: increase assert_screen timeout in wait_for_desktop

d3flex:
Give more time to resolve needle match failure after reboot

Issue: https://progress.opensuse.org/issues/199217

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -43,7 +43,7 @@ sub handle_gui_password {
 }
 
 sub wait_for_desktop {
-    assert_screen([qw/boot-menu openqa-desktop/]);
+    assert_screen([qw/boot-menu openqa-desktop/], 500);
     send_key 'ret' if match_has_tag('boot-menu');
     assert_screen 'openqa-desktop', 500;
     return if match_has_tag('generic-desktop');


d3flex:
https://openqa.opensuse.org/tests/overview?build=%3Aiob&distri=openqa&version=Tumbleweed look for only vr_apparmor_iob_*. at first 100 iterations there are some jobs restarted. the reason was irrelevant - not because apparmor. a second iteration is in progress at the time of writing

Martchus (dismissed):
I actually ran into this recently when trying to run tests locally which also sometimes failed on that match.

perlpunk (changes requested):
I'm don't know how waiting 8.3 minutes will help, if the openqa-desktop is already shortly visible and then switches to the console.
I would like to see enough verification runs showing that it would switch back to the openqa-desktop after switching to the console.

okurz (changes requested):
please stick to https://progress.opensuse.org/issues/199217#Suggestions first. The first task this mentions:
> Collect logs so we can investigate why GUI is crashing or not starting after apparmor is enabled

d3flex:
> I'm don't know how waiting 8.3 minutes will help, if the openqa-desktop is already shortly visible and then switches to the console. I would like to see enough verification runs showing that it would switch back to the openqa-desktop after switching to the console.

I used the same timeout as `assert_screen 'openqa-desktop', 500;`. I run 100x2 in o3 - no failure on apparmor. I do not follow the "it would switch back to the openqa-desktop after switching to the console.". Likely is something that you notice but the goal is to make the module not produce false negatives, right?

d3flex:
> please stick to https://progress.opensuse.org/issues/199217#Suggestions first. The first task this mentions:
> 
> > Collect logs so we can investigate why GUI is crashing or not starting after apparmor is enabled

I did not encounter any crashes. The commit is not wrong in any significant way and seems to improve the test. It can be merged, resolve the alert ticket and move on. if there is a crash we can come back to it.

okurz:
> The commit is not wrong in any significant way and seems to improve the test.

Please provide a proof for this statement. If you state that a sporadic issue is gone or improved by your change then please provide the statistical analyis showing the fail-ratio before and after your change to prove the improvement.

d3flex:
> > The commit is not wrong in any significant way and seems to improve the test.
> 
> Please provide a proof for this statement. If you state that a sporadic issue is gone or improved by your change then please provide the statistical analyis showing the fail-ratio before and after your change to prove the improvement.

it is already provided 
https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/284#issuecomment-4257817441

But I do not understand whats the problem with this PR. https://progress.opensuse.org/issues/199217 is high priority. Why do we have to keep this open even if there is more investigation to be done?

perlpunk:
> I do not follow the "it would switch back to the openqa-desktop after switching to the console.". Likely is something that you notice but the goal is to make the module not produce false negatives, right?

I noticed it and wrote a comment in the ticket. Didn't you see the comment?
https://progress.opensuse.org/issues/199217#note-3
> There is actually a short time when the X11 login appears. Go to https://openqa.opensuse.org/tests/5834014/video?filename=video.webm&t=24.83,24.88 and play.
> But then the console login appears.

Did you watch the video?

But if you are sure you fixed this by waiting longer that would mean it switches from the desktop to the console and then (after a long while) back again.
The default timeout is 30s. I think 500s is excessive. That would be a really weird behaviour and should still be investigated, if it takes that long.

d3flex:
> > I do not follow the "it would switch back to the openqa-desktop after switching to the console.". Likely is something that you notice but the goal is to make the module not produce false negatives, right?
> 
> I noticed it and wrote a comment in the ticket. Didn't you see the comment? https://progress.opensuse.org/issues/199217#note-3
> 
> > There is actually a short time when the X11 login appears. Go to https://openqa.opensuse.org/tests/5834014/video?filename=video.webm&t=24.83,24.88 and play.
> > But then the console login appears.
> 
> Did you watch the video?
> 
> But if you are sure you fixed this by waiting longer that would mean it switches from the desktop to the console and then (after a long while) back again. The default timeout is 30s. I think 500s is excessive. That would be a really weird behaviour and should still be investigated, if it takes that long.

I did just now. and then I tried to find it in one of the VR. I didnt see it. I agree it looks weird watching the video and it might be something additional on it.
I just triggered another series of tests

perlpunk:
I think now I understand it. The desktop appears in the video, but only slightly after 30 seconds (judging from the timestamps in the video), which is just a tiny bit too late for the default timeout of 30 seconds.
So with this I would suggest to bump the timeout to 60 seconds.
Since this is called in different locations, maybe the `wait_for_desktop` can get a new timeout parameter, since the reboot seems to be the only place where it takes that long.
E.g. by default we call `wait_for_desktop`, and here we would call `wait_for_desktop(timeout => 60)`.

perlpunk:
Looking at the videos where the test is passing, it usually takes something like 25-28 seconds until the desktop appears, so with that I think 60s would be a good timeout here.

okurz (approved):

okurz (changes requested):
Sorry

lib/utils.pm
```diff
@@ -43,7 +43,7 @@ sub handle_gui_password {
 }
 
 sub wait_for_desktop {
-    assert_screen([qw/boot-menu openqa-desktop/]);
+    assert_screen([qw/boot-menu openqa-desktop/], 500);
```
perlpunk, line 46:
```perl
sub wait_for_desktop (%args) {
    assert_screen([qw/boot-menu openqa-desktop/], $args{timeout});
```
and add `60` as an argument in the apparmor module.

perlpunk (dismissed):
LIke I suggested earlier in comments

d3flex:
> Looking at the videos where the test is passing, it usually takes something like 25-28 seconds until the desktop appears, so with that I think 60s would be a good timeout here.

As I mentioned, I just took the same value from the assert_screen below - if you think it is worth to descrease it, I can try but wouldnt be akward that one has 60 and the other 500? (I mean even if it doesnt make sense, we are quite tolerant in timeout/retries usually in the sake of safety or stability)

perlpunk:
I don't know why the other assert_screen has 500, maybe worth to decrease it.
If we use 500 here as well, it means that in case of failure sit will take 8.3 minutes until the job finishes.

I give up, I found out what was actually happening, I made a suggestion, now others can continue the discussion.

okurz (approved):
Actually I think we can improve the test code to not re-check the desktop twice. I will see what I can do on top. Until then 500 is fine

Martchus (approved):
Then let's merge it after all (considering @okurz statement).

PR #285, opened 2026-04-16
Login before collecting logs on failure

kalikiana:
Otherwise in some cases no logs are available and the post_fail_hook times out.

See: https://progress.opensuse.org/issues/199217

### DIFF
--- a/lib/openQAcoretest.pm
+++ b/lib/openQAcoretest.pm
@@ -1,10 +1,13 @@
 package openQAcoretest;
 use Mojo::Base 'basetest', -signatures;
 use testapi;
-use utils qw(switch_to_root_console get_log);
+use utils qw(login switch_to_root_console get_log);
 
 sub post_fail_hook ($self) {
     switch_to_root_console;
+    # we may not be logged in yet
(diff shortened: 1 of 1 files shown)


lib/openQAcoretest.pm
```diff
@@ -1,10 +1,13 @@
...
 use Mojo::Base 'basetest', -signatures;
 use testapi;
-use utils qw(switch_to_root_console get_log);
+use utils qw(login switch_to_root_console get_log);
 
 sub post_fail_hook ($self) {
     switch_to_root_console;
+    # we may not be logged in yet
+    assert_screen([qw/root-console inst-console/]);
+    login if match_has_tag('inst-console');
```
okurz, line 10: IMHO this is going to far with workarounds as this test distribution never implemented the proper way to call `select_console` which should ensure that we switch to the right console and are actually logged in. Please look into using that. Look into https://open.qa/api/testapi/#_add_console and https://open.qa/api/testapi/#_select_console

okurz (changes requested):

kalikiana:
Closing since the ticket was resolved and I guess we don't care about the logs anymore

PR #288, opened 2026-04-17
refactor: simplify wait_for_desktop

d3flex:
- remove duplicate assert_screen
- reduce timeout

Issue: https://progress.opensuse.org/issues/199217

### DIFF
--- a/lib/utils.pm
+++ b/lib/utils.pm
@@ -43,9 +43,8 @@ sub handle_gui_password {
 }
 
 sub wait_for_desktop {
-    assert_screen([qw/boot-menu openqa-desktop/], 500);
+    assert_screen([qw/boot-menu openqa-desktop/], 200);
     send_key 'ret' if match_has_tag('boot-menu');
-    assert_screen 'openqa-desktop', 500;
     return if match_has_tag('generic-desktop');
     if (match_has_tag('openqa-desktop-locked')) {
         send_key 'esc';


d3flex:
All openqa_install_nginx-vr_new_iob_* jobs out of 200 pass in https://openqa.opensuse.org/tests/overview?version=Tumbleweed&distri=openqa&build=%3Aiob

lib/utils.pm
```diff
@@ -43,9 +43,8 @@ sub handle_gui_password {
 }
 
 sub wait_for_desktop {
-    assert_screen([qw/boot-menu openqa-desktop/], 500);
+    assert_screen([qw/boot-menu openqa-desktop/], 200);
     send_key 'ret' if match_has_tag('boot-menu');
-    assert_screen 'openqa-desktop', 500;
```
okurz, line 48: That's wrong. If we have a "boot-menu" then the next lines would return true and we would end up on `handle_gui_password` which would call `assert_screen 'lockscreen-password-prompt';` but we would need to handle the login, not the unlock.

okurz (changes requested):
Please see https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/287

lib/utils.pm
kalikiana: Indeed. Which reminds me that I wanted to [extend the docs to explain this behavior](https://github.com/os-autoinst/os-autoinst/pull/2882).

d3flex:
improved fix by https://github.com/os-autoinst/os-autoinst-distri-openQA/pull/287

PR #289, opened 2026-04-28
fix: require SUT reboot for NetworkManager in multi-machine tests

gauravpathak:
Ticket: https://progress.opensuse.org/issues/199742  
Validation Test: https://openqa.opensuse.org/tests/5881635#step/openqa_worker/23  

Updates the worker setup process to conditionally reboot the SUT when NetworkManager is in use, ensuring the network configurations applied by `os-autoinst-setup-multimachine` take effect properly.

### DIFF
--- a/tests/install/openqa_worker.pm
+++ b/tests/install/openqa_worker.pm
@@ -30,6 +30,11 @@ EOF
     assert_script_run "systemctl enable --now openqa-worker@2" if get_var('FULL_MM_TEST');
     save_screenshot;
     clear_root_console;
+    if (script_run('systemctl is-active NetworkManager.service') == 0) {
+        assert_script_run('systemctl reboot');
+        wait_for_desktop;
+        login;
+    }
 }
 
 1;


gauravpathak:
More Validation Test Runs: https://openqa.opensuse.org/tests/overview?distri=openqa&version=Tumbleweed&build=%3Agpathak-oqa-in-oqa

Martchus (approved):
I'm wondering why restarting `NetworkManager.service` and perhaps also `openvswitch.service` is not sufficient but I guess this change is ok.

okurz (changes requested):
IMHO this is doing way too much in a test. The setup should ensure a working environment, not the tests.

gauravpathak:
> IMHO this is doing way too much in a test. The setup should ensure a working environment, not the tests.

I tried various possibilities in `os-autoinst-setup-multi-machine` itself including these as well https://github.com/os-autoinst/os-autoinst/compare/master...gauravpathak:os-autoinst:mmsetup-nm but the multi-machine tests are still failing for openQA-in-openQA.  
We also rebooted all our OSD worker hosts when we upgraded and migrated them from wicked to NetworkManager.

PR #290, opened 2026-04-28
fix: ensure br1 for cleanup and prevent timeout from get_logs

d3flex:
Skip `ip a del` on br1 when the bridge doesn't exist, preventing unnecessary errors in test_running and post_fail_hook. Increase get_log timeout to 300s to prevent journalctl dumps from timing out.


Martchus (approved):

gauravpathak (approved):

okurz (changes requested):
s/timout/timeout/

okurz (approved):

PR #291, opened 2026-05-19
fix: skip unnecessary login after switch to 'None' provider

okurz:
Related issue: https://progress.opensuse.org/issues/194786

### DIFF
--- a/main.pm
+++ b/main.pm
@@ -45,7 +45,7 @@ sub load_openQA_tests() {
     }
     else {
       loadtest 'openQA/dashboard';
-      loadtest 'openQA/login';
+      loadtest 'openQA/login' unless get_var('OPENQA_FROM_BOOTSTRAP');
       # testing from git does not schedule tests so far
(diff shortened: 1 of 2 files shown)


main.pm
```diff
@@ -45,7 +45,7 @@ sub load_openQA_tests() {
     }
     else {
       loadtest 'openQA/dashboard';
-      loadtest 'openQA/login';
+      loadtest 'openQA/login' unless get_var('OPENQA_NEEDS_LOGIN');
```
perlpunk: Where is this variable set?
The logic sounds opposite to the name of the variable to me. If OPENQA_NEEDS_LOGIN, then we need to load the login test.

okurz: Correct. Wrong way around

kalikiana (approved):

nicksinger (approved):

perlpunk:
openqa_from_bootstrap https://openqa.opensuse.org/tests/5943821#step/login/3 is now failing

PR #295, opened 2026-06-01
feat: use proper "select_console" calls

okurz:
Instead of manually switching tty's with no consistent synchronisation
we should use the "proper" os-autoinst provided way to define
consoles with "add_console" and use them with "select_console". This
change simplifies code, clarifies the intention and prevents
synchronisation issues when actions are conducted before the
availability of fully logged in terminals was ensured.

### DIFF
--- a/lib/susedistribution.pm
+++ b/lib/susedistribution.pm
@@ -1,6 +1,16 @@
 package susedistribution;
-use base 'distribution';
+use Mojo::Base 'distribution', -signatures;
+use File::Basename qw(basename);
+use testapi qw(get_required_var check_var);
 
-# Base class for all test modules
+sub init ($self) {
+    $self->SUPER::init();
+    $self->add_console('root-console', 'tty-console', {tty => 3});
+    my @hdd = split(/-/, basename get_required_var('HDD_1'));
+    # older openSUSE Tumbleweed has x11 still on tty7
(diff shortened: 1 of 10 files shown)


perlpunk (approved):

gauravpathak (approved):

lib/susedistribution.pm
```diff
@@ -1,6 +1,16 @@
...
 
-# Base class for all test modules
+sub init ($self) {
+    $self->SUPER::init();
+    $self->add_console('root-console', 'tty-console', {tty => 3});
+    my @hdd = split(/-/, basename get_required_var('HDD_1'));
+    # older openSUSE Tumbleweed has x11 still on tty7
+    # minimalx has x11 on tty7
+    $self->add_console('x11', 'tty-console', {tty => ($hdd[3] < 20190617 or check_var('DESKTOP', 'minimalx')) ? 7 : 2});
+;
```
r-richardson, line 13: i know its already merged, but this semicolon looks redundant

perlpunk: We should probably enable perltidy and perlcritic in this repo

okurz: @r-richardson why not open a PR to fix that?

perlpunk: * #297

PR #297, opened 2026-06-16
Add style checks and os-autoinst-common

perlpunk:


okurz (approved):

.mergify.yml
```diff
@@ -2,26 +2,26 @@ pull_request_rules:
...
+          - linear-history
       - and:
-        - "#approved-reviews-by>=2"
-        - "#changes-requested-reviews-by=0"
-        # https://doc.mergify.io/examples.html#require-all-requested-reviews-to-be-approved
-        - "#review-requested=0"
+          - "#approved-reviews-by>=2"
+          - "#changes-requested-reviews-by=0"
+          # https://doc.mergify.io/examples.html#require-all-requested-reviews-to-be-approved
+          - "#review-requested=0"
```
Martchus, line 15: That reminds me of a question I had when investigating the auto-merging problems earlier. Why do we use more than one `and` here? Below we even have:

> mergify config checks needs at least two rules in "and" so we repeat …

Couldn't that be completely avoided?

(Of course we shouldn't restructure this as part of a style change.)

Martchus (approved):

.mergify.yml
perlpunk: Probably because that's the only way to use aliases like `&base_checks`

PR #300, opened 2026-07-07
feat(worker_ay_validation): use wget internal retry on host error

okurz:
Same as done in osautoinst/start_test.pm where we also use wget.


perlpunk (approved):

okurz:
shortened the git commit message subject line to cover the conventional message check

Martchus (approved):

PR #305, opened 2026-07-28
fix: Work around libsvtav1 crashing in nested test

Martchus:
After the SVT-AV1 encoder lib was updated in TW from v3.0.0 to v4.2.0 it crashes when executed in our virtualized test environment. (This can be reproduced by cloning a job of flavor `dev-full`. It does not crash when ffmpeg/libsvtav1 is executed on a bare metal host. Hence, when cloning on a TW host, only the inner test fails.)

It helps to set the level of parallelism explicitly. So this change sets it explicitly to 1 via `-svtav1-params lp=1` which is in accordance with what is otherwise automatically detected inside our test VM. So this change should not make any difference but apparently this way libstav1 no longer crashes.

Note that these crashes are reproducible 100 % of the time and without specifying the level of parallelism. When specifying it (the exact level being specified does not matter) I could not reproduce one crash so far.

Note that using `-svtav1-params asm=0` does not make a difference. (So it still crashes with `asm level selected : up to c`.)

Related ticket: https://progress.opensuse.org/issues/204756

### DIFF
--- a/tests/install/openqa_worker.pm
+++ b/tests/install/openqa_worker.pm
@@ -2,6 +2,12 @@ use Mojo::Base 'openQAcoretest';
 use testapi;
 use utils;
 
+# work around libsvtav1 crashing in the VM without `-svtav1-params lp=…`
+# see https://progress.opensuse.org/issues/204756#note-5 and https://gitlab.com/AOMediaCodec/SVT-AV1/-/work_items/2383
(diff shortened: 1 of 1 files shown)


perlpunk (approved):

okurz (changes requested):
> So this change should not make any difference but apparently this way libstav1 no longer crashes.

please elaborate on the "apparently". Do you have statistics proofing that? I am also fine to get those statistics from production after merge and then consider to revert again if that is your plan

okurz (approved):

PR #308, opened 2026-08-14
feat: Avoid needless logout and re-login

Martchus:
* Leave a clean root console in the `test_distribution` test module
* Remove the error prone code in the `worker` test module to re-login (it is error prone as it lacks `assert_screen 'inst-console';` as done in `utils::login`)
* Note that the `dashboard` test module already has `select_console 'x11';` so there is not need for `test_distribution` doing the same

Related ticket: https://progress.opensuse.org/issues/205527

### DIFF
--- a/tests/install/test_distribution.pm
+++ b/tests/install/test_distribution.pm
@@ -15,10 +15,8 @@ sub run {
     # /etc/zypp/zypp.conf so we need to refresh explicitly with retries in
     # case of problems.
     install_packages('os-autoinst-distri-opensuse-deps');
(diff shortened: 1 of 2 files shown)


okurz (approved):

perlpunk (approved):

okurz:
@Martchus just in case you don't know (anymore?) but openQA tests are automatically triggered and https://openqa.opensuse.org/tests/6163804#step/worker/1 looks good so unless you want to change the test code you can just undraft.

Martchus:
> @Martchus just in case you don't know (anymore?)

I know, that's why I created this draft instead of running tests locally :-)

